PR #23, opened 2021-05-20
Fixed regex for THM, HTB and CTF flags

amadejpapez:
(?i) can only be at the start of the regex. This caused warnings when running test


bee-san:
Thank youu!!!!!!!!!!!!!!!!!!!

PR #26, opened 2021-05-21
Regex for more crypto wallets, IPv6 and improved coordinates

amadejpapez:
- pyWhat can now detect Ripple, Bitcoin Cash, Litecoin and Monero!
- added detection for IPv6
- if there is a port number alongside of the IP address, it will now grab that too
- improved detection for coordinates, added formats (N 32° 53.733 W 096° 48.358") and made it less restrictive (20,80 is no longer a coordinate :D)

PR #36, opened 2021-05-23
Regex for LTC, BCH, XRP, XMR crypto wallets, IPv6 and more

### LINKED ISSUE #33: YouTube Video ID Regex too broad, fails on SSN
Regex fails if we use the SSN from : https://www.ssa.gov/history/ssn/misused.html

```
078-05-1120
```

It matches as a youtube video-id first. Ideally we:
* Move SSN physically up the JSON file
So it matches on SSN first?

Alternatively a #29 filter system but I think physically moving the regex is fine?

amadejpapez:
- pyWhat can now detect Ripple, Bitcoin Cash, Litecoin and Monero wallet addresses!
- added regex for IPv6
- if there is a port number alongside of the IP address, it will now grab that too
- improved detection for coordinates, added formats like `N 32° 53.733 W 096° 48.358"` and made it less restrictive (`20,80` is no longer a coordinate)
- added a much better URL regex from @SkeletalDemise
- URL regex is now also case insensitive (before URL starting with `HTTP` would not match)
- made YouTube id regex more restrictive: now it requires at least one number, at least one uppercase letter and at least one lowercase letter (`078-05-1120` no longer matches, but `078-05-1aA0` does), should fix #33 issue
- and with all of the changes requested in #26 :D

### DIFF
--- a/pywhat/Data/regex.json
+++ b/pywhat/Data/regex.json
@@ -1,7 +1,7 @@
 [
     {
-       "Name": "Ethereum Wallet Address",
-       "Regex": "^0x[a-fA-F0-9]{40}$",
+       "Name": "Ethereum (ETH) Wallet Address",
+       "Regex": "(?i)^0x[a-f0-9]{40}$",
        "Description": null,
        "URL": "https://etherscan.io/address/",
        "Rarity": 1,
@@ -12,7 +12,7 @@
     },
     {
        "Name": "Access-Control-Allow-Header",
-       "Regex": "Access-Control-Allow: [a-zA-Z0-9\\-*]",
+       "Regex": "(?i)Access-Control-Allow: [a-z0-9\\-*]",
        "Description": "Used for [#CAE4F1][link=https://en.wikipedia.org/wiki/Cross-origin_resource_sharing]Cross-Origin Resource Sharing (CORS)[/link][/#CAE4F1]",
        "Rarity": 1,
        "Tags": [
@@ -21,7 +21,7 @@
       ]
     },
     {
-       "Name": "Bitcoin Wallet",
+       "Name": "Bitcoin (₿) Wallet Address",
        "Regex": "^[13][a-km-zA-HJ-NP-Z1-9]{25,34}$",
        "Description": null,
        "URL": "https://www.blockchain.com/btc/address/",
@@ -69,7 +69,7 @@
     },
     {
        "Name": "Uniform Resource Locator (URL)",
-       "Regex": "(https?:\\/\\/(?:www\\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\\.[^\\s]{2,}|www\\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\\.[^\\s]{2,}|https?:\\/\\/(?:www\\.|(?!www))[a-zA-Z0-9]+\\.[^\\s]{2,}|www\\.[a-zA-Z0-9]+\\.[^\\s]{2,})",
+       "Regex": "(?i)^(?:(?:(?:https?|ftp):)?\/\/)(?:\\S+(?::\\S*)?@?)?(?:(?!(?:10|127)(?:\\.\\d{1,3}){3})(?!(?:169\\.254|192\\.168)(?:\\.\\d{1,3}){2})(?!172\\.(?:1[6-9]|2\\d|3[0-1])(?:\\.\\d{1,3}){2})(?:[1-9]\\d?|1\\d\\d|2[01]\\d|22[0-3])(?:\\.(?:1?\\d{1,2}|2[0-4]\\d|25[0-5])){2}(?:\\.(?:[1-9]\\d?|1\\d\\d|2[0-4]\\d|25[0-4]))|(?:(?:[a-z0-9\u00a1-\uffff][a-z0-9\u00a1-\uffff_-]{0,62})?[a-z0-9\u00a1-\uffff]?\\.?)+?(?:[a-z\u00a1-\uffff]+\\.?))(?::\\d{2,5})?(?:[\/?#]\\S*)?$",
        "Description": null,
        "Rarity": 1,
        "Tags": [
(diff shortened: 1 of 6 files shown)


pywhat/Data/regex.json
```diff
@@ -12,7 +12,7 @@
     },
     {
        "Name": "Access-Control-Allow-Header",
-       "Regex": "Access-Control-Allow: [a-zA-Z0-9\\-*]",
+       "Regex": "(?i)Access-Control-Allow: [a-z0-9\\-*]",
```
bee-san, line 15: Good spot!

pywhat/Data/regex.json
```diff
@@ -69,26 +69,48 @@
     },
     {
        "Name": "Uniform Resource Locator (URL)",
-       "Regex": "(https?:\\/\\/(?:www\\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\\.[^\\s]{2,}|www\\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\\.[^\\s]{2,}|https?:\\/\\/(?:www\\.|(?!www))[a-zA-Z0-9]+\\.[^\\s]{2,}|www\\.[a-zA-Z0-9]+\\.[^\\s]{2,})",
+       "Regex": "(?i)^(?:(?:(?:https?|ftp):)?\/\/)(?:\\S+(?::\\S*)?@?)?(?:(?!(?:10|127)(?:\\.\\d{1,3}){3})(?!(?:169\\.254|192\\.168)(?:\\.\\d{1,3}){2})(?!172\\.(?:1[6-9]|2\\d|3[0-1])(?:\\.\\d{1,3}){2})(?:[1-9]\\d?|1\\d\\d|2[01]\\d|22[0-3])(?:\\.(?:1?\\d{1,2}|2[0-4]\\d|25[0-5])){2}(?:\\.(?:[1-9]\\d?|1\\d\\d|2[0-4]\\d|25[0-4]))|(?:(?:[a-z0-9\u00a1-\uffff][a-z0-9\u00a1-\uffff_-]{0,62})?[a-z0-9\u00a1-\uffff]?\\.?)+?(?:[a-z\u00a1-\uffff]+\\.?))(?::\\d{2,5})?(?:[\/?#]\\S*)?$",
        "Description": null,
        "Rarity": 1,
```
bee-san, line 74: Is it really 1? What's the most unnatural domain that can pass this? :-) Maybe we need a lot of tests to be sure? <3

pywhat/Data/regex.json
```diff
@@ -69,26 +69,48 @@
...
+       "Rarity": 1,
+       "Tags": [
+          "Cryptocurrency",
+          "Finance"
+       ]
+    },
+    {
+       "Name": "Internet Protocol (IP) Address Version 6",
+       "Regex": "\\[?(?:(?:[0-9a-f]{1,4}:){7,7}[0-9a-f]{1,4}|([0-9a-f]{4}:){1,7}:|([0-9a-f]{1,4}:){1,6}:[0-9a-f]{1,4}|([0-9a-f]{1,4}:){1,5}(:[0-9a-f]{1,4}){1,2}|([0-9a-f]{1,4}:){1,4}(:[0-9a-f]{1,4}){1,3}|([0-9a-f]{1,4}:){1,3}(:[0-9a-f]{1,4}){1,4}|([0-9a-fA]{1,4}:){1,2}(:[0-9a-f]{1,4}){1,5}|[0-9a-f]{1,4}:((:[0-9a-f]{1,4}){1,6})|:((:[0-9a-f]{1,4}){1,7}|:)|fe80:(:[0-9a-f]{0,4}){0,4}%[0-9a-z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|(?:[0-9a-f]{1,4}:){1,4}:(?:(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(?:2[0-4]|1{0,1}[0-9]){0,1}[0-9]))\\]?(?::[0-9]{1,5})?",
+       "Description": "Universal address pointing to an entity on the web.",
```
bee-san, line 93: Remove this description please! We only want descriptions on things that are naturally confusing (Like Social Security Numbers for non-americans) :-)

pywhat/Data/regex.json
```diff
@@ -69,26 +69,48 @@
...
+       "URL": "https://www.shodan.io/host/",
+       "Rarity": 1,
+       "Tags": [
+          "Identifiers",
+          "Networking"
+       ]
+    },
+    {
+       "Name":"Latitude & Longitude Coordinates",
+       "Regex":"(?i)^(?:(?:N|W|S|E)\\s?\\d+\\s?\\u00B0?\\s?\\d+\\.?\\d*\\s?\\'?\\s?\\d*\\.?\\,?\\d*?\"?\\s?){1,2}$|^(?:\\d+\\s?\\u00B0\\s?\\d+\\s?\\'\\s?\\d+\\.?\\,?\\d{0,}?\"\\s?(?:N|W|S|E)\\s?){1,2}$|^(?:[-+]?(?:[0-8]?\\d+\\.\\d{4,}|90(?:\\.0+)?),\\s*[-+]?(?:180(?:\\.0+)?|(?:(?:1[0-7]\\d)|(?:[1-9]?\\d))(?:\\.\\d+)?))$",
```
bee-san, line 103: This is hard-core! Well done!

pywhat/Data/regex.json
```diff
@@ -239,7 +294,7 @@
     },
     {
        "Name": "Email Address",
-       "Regex": "(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$)",
+       "Regex": "(?i)(^[a-z0-9_.+-]+@[a-z0-9-]+\\.[a-z0-9-.]+$)",
```
bee-san, line 297: Email addresses are super weird, see this regex (maybe make a separate PR for this! This is **huge**!)
https://stackoverflow.com/questions/201323/how-to-validate-an-email-address-using-a-regular-expression
Also make sure that regex isn't super slow :-)

pywhat/Data/regex.json
```diff
@@ -249,10 +304,10 @@
     },
     {
        "Name": "YouTube Video ID",
-       "Regex": "^[0-9A-Za-z_-]{10}[048AEIMQUYcgkosw]{1}$",
+       "Regex": "^(?=.*[0-9])(?=.*[A-Z])(?=.*[a-z])[0-9A-Za-z_-]{10}[048AEIMQUYcgkosw]{1}$",
```
bee-san, line 307: Fun fact: 

I clicked a bunch of videos on my homepage, the 3rd video's URL was this:
https://www.youtube.com/watch?v=zKOMLoBP_Xw

so `_` does appear!

I would drop this down to a 0.2 personally -- it keeps on showing up and is annoying 😄 

_Perhaps_ we can implement some checkers (in the future!) that run Luhn checks (For cards) and sees if the youtube video exists (see https://www.youtube.com/watch?v=dlkadkjsak )

But for now, just drop it 😄

pywhat/Data/regex.json
```diff
@@ -268,9 +323,10 @@
        ]
     },
     {
-       "Name": "Internet Protocol (IP) Address",
-       "Regex": "(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)",
+       "Name": "Internet Protocol (IP) Address Version 4",
+       "Regex": "(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(?::[0-9]{1,5})?",
        "Description": null,
+       "URL": "https://www.shodan.io/host/",
        "Rarity": 0.4,
```
bee-san, line 330: This should be higher than 0.4. I feel like :)

tests/test_regex_identifier.py
```diff
@@ -39,19 +39,54 @@ def test_lat_long2():
     assert "Latitude & Longitude Coordinates" in res[0]["Regex Pattern"]["Name"]
 
 
+def test_lat_long3():
+    r = regex_identifier.RegexIdentifier()
+    res = r.check(['77\u00B0 30\' 29.9988" N'])
```
bee-san, line 44: Whats `\u00B0`? Maybe a comment explaining?

tests/test_regex_identifier.py
```diff
@@ -39,19 +39,54 @@ def test_lat_long2():
...
+    r = regex_identifier.RegexIdentifier()
+    res = r.check(["40.741895,-73.989308"])
+    assert "Latitude & Longitude Coordinates" in res[0]["Regex Pattern"]["Name"]
+
+
 def test_ip():
     r = regex_identifier.RegexIdentifier()
     res = r.check(["http://10.1.1.1"])
-    assert "Uniform Resource Locator (URL)" in res[0]["Regex Pattern"]["Name"]
+    assert "Internet Protocol (IP) Address Version 4" in res[0]["Regex Pattern"]["Name"]
```
bee-san, line 69: You can also do the reverse, assert "URL" not in res? Be aware that this:
```py
res[0]
```

only gets the first, most likely output. If you want to make sure this doesn't match on domains you'll need to check all of em :-)

tests/test_regex_identifier.py
```diff
@@ -39,19 +39,54 @@ def test_lat_long2():
...
+
+
+def test_ip4():
+    r = regex_identifier.RegexIdentifier()
+    res = r.check(["[2001:db8::1]:8080"])
+    assert "[2001:db8::1]:8080" in res[0]["Matched"]
 
 
-@pytest.mark.skip(reason="Fails Regex due to http://")
 def test_internationak_url():
```
bee-san, line 90: woooooooooooooooooooo we can stop skipping tests!!!

pywhat/Data/regex.json
bee-san: PS: We can use that video as a test :eyes:

bee-san (changes requested):
Great work!!!!!!!!!!!!!!!!!!!!!!!!!!!! This is gonna be so good :fire:

pywhat/Data/regex.json
bee-san: Something we should see with testing, if we test well enough -- if we drop this down to 0.2 does this break the assumptions in our tests? Does the program still work as expected, minus those glitches?

Also, can you make a test for this glitch? Anytime someone says "XYZ doesn't work" and you fix it, make a test.

So test this:
```
078-05-1120
```

And test to make sure youtube videos don't appear first or above SSNs by doing `ret[0]`.

That way our tests form around what we expect our program to do, and if we change something (like moving this regex up or down the popularity ratings) will it break the program? Will it break what we expected it to do?

bee-san: ^^ This is also why I have tests that don't run and are skipped, it's better to have tests that don't run and skip then to have no tests at all. That way we know where the limits are and can work on fixing them! :-)

tests/test_regex_identifier.py
amadejpapez: It's a degree symbol (°) as otherwise Windows tests fail. Will write a comment in test for that

bee-san (approved):

noona88:
> - pyWhat can now detect Ripple, Bitcoin Cash, Litecoin and Monero wallet addresses!
> 
> - added regex for IPv6
> 
> - if there is a port number alongside of the IP address, it will now grab that too
> 
> - improved detection for coordinates, added formats like `N 32° 53.733 W 096° 48.358"` and made it less restrictive (`20,80` is no longer a coordinate)
> 
> - made YouTube id regex more restrictive: now it requires at least one number, at least one uppercase letter and at least one lowercase letter (`078-05-1120` no longer matches, but `078-05-1aA0` does), should fix #33 issue
> 
> - added a much better URL regex from @SkeletalDemise
> 
> - URL regex is now also case insensitive (before URL starting with `HTTP` would not match)
> 
> - and with all of the changes requested in #26 :D

PR #41, opened 2021-05-25
Improved detection for emails

amadejpapez:
- new email regex so some of the *weird* emails will also be detected!
  - emails with any of the `!#$%&'*+-/=?^_{|}~` symbols will now match so this regex should match everything that is considered a valid email address
  - emails like `email@[123.123.123.123]` and `'email'@example.com` will now also match
- removed `URL` from Monero wallet address as it should not be there in the first place :)


bee-san (approved):
As always, great testing! 😄

PR #54, opened 2021-05-28
Minor fixes

piatrashkakanstantinass:


pywhat/__main__.py
```diff
@@ -4,21 +4,16 @@
 pywhat: Identify Anything.
 """
 
+import platform
 import sys
 
 if __name__ == "__main__":
     major = sys.version_info[0]
     minor = sys.version_info[1]
```
bee-san, line 12: Instead of using `sys` here, we can use `python_version`?

> platform.python_version()
> Returns the Python version as string 'major.minor.patchlevel'.
> 
> Note that unlike the Python sys.version, the returned value will always include the patchlevel (it defaults to 0).

So to get major and minor, we can do:

```py
python_version = platform.python_version()
versions = python_version.split(".")
major = versions[0]
minor = versions[1]
```

This saves us from importing `sys` :-)

pywhat/__main__.py
```diff
@@ -4,21 +4,16 @@
...
-        str(sys.version_info[0])
-        + "."
-        + str(sys.version_info[1])
-        + "."
-        + str(sys.version_info[2])
-    )
+    python_version = platform.python_version()
 
-    if major != 3 or (major == 3 and minor < 6):
+    if major < 3 or (major == 3 and minor < 6):
```
bee-san, line 16: Good spot!!! It should be `<` here 😄

bee-san (changes requested):

bee-san:
Thank you for the PR! 🙌🏻

pywhat/__main__.py
piatrashkakanstantinass: Going to change that!

piatrashkakanstantinass: @bee-san, you forgot that sys.argv and sys.exit() are being used. Therefore, I think that these changes are not necessary.

bee-san (approved):

Issue #43, opened 2021-05-26
Using a folder path as input


Jayy001:
* What is a great tool for identifying anything, but it also supports file input such as .pcap
* Sometimes, we have loads of of files we want to scan for but it will take a long time to manually input each file into what

This is where the "folder" input comes in, give it a folder path and for each file / folder in there it will identify it

zarszz:
it look interesting, it like give folder/directory as input and application scan files in entire folder right ?

PR #58, opened 2021-05-30
Added recursion support

amadejpapez:
Want to search the whole folder and see what all of the files in that folder contain? pyWhat can now do exactly that!

- if input is a directory, recursively search through it (#43)
- now that pyWhat is searching for multiple files at once, if a file is identified with magic numbers, it will also print the name of the file, so it will be easier to see which one is identified
- it pyWhat is printing matches from multiple files, it will also print in which file each match is located in a new `Filename` row (if input is only text, this row is not there)
- API now by default also checks if input is a file or a folder. If you want to disable this and only treat input as text, add `True` to `identify()` like this `identify("file.txt", True)`
- hidden file found with recursion has a CTF flag in the name? pyWhat is now also searching for matches inside of filenames!
- added recursion support to the README and to `--help` command
- added links to all of the badges in README (#60)
- Discover Number can no longer be matched inside of a string


bee-san:
Can I get a screenshot to see what this looks like, pls? 💖

amadejpapez:
> Can I get a screenshot to see what this looks like, pls?

image

bee-san:
Do you think it'd be better if we printed what file it came from in the table as a new column?

amadejpapez:
> Do you think it'd be better if we printed what file it came from in the table as a new column?

Oo that's a good idea but I don't there is enough space to add another column. Maybe a separate table for each file would be better?

bee-san:
> > Do you think it'd be better if we printed what file it came from in the table as a new column?
> 
> Oo that's a good idea but I don't there is enough space to add another column. Maybe a separate table for each file would be better?

> I don't there is enough space to add another column
Why not? Most lines continue onto the next line if its too long? 😄

amadejpapez:
> Why not? Most lines continue onto the next line if its too long?

Yeah you are right, there is enough space :grin: 

image

pywhat/identifier.py
```diff
@@ -12,24 +12,42 @@ def __init__(self):
 
     def identify(self, text: str, api=False) -> dict:
         identify_obj = {}
+        magic_numbers = {}
+        file_text = []
+
+        if not api and (os.path.isdir(text) == True or os.path.isfile(text) == True):
```
piatrashkakanstantinass, line 18: You can use `if not api and os.path.exists(text):`. The code will be shorter.

amadejpapez: Good idea! Will change it :)

bee-san, line 18: What should the behaviour of the API be _if_ it was given a file path?

pywhat/printer.py
```diff
@@ -14,12 +15,13 @@ def pretty_print(self, text: dict):
...
-            to_out += "\n"
-        if to_out:
-            console.print(to_out)
+            for key, value in text["File Signatures"].items():
+                if value:
+                    key = os.path.basename(key)
+                    to_out += "\n"
+                    to_out += f"[bold #D7Afff]File Identified[/bold #D7Afff]: [bold]{key}[/bold] with Magic Numbers {value['ISO 8859-1']}."
+                    to_out += f"\n[bold #D7Afff]File Description:[/bold #D7Afff] {value['Description']}."
+                    to_out += "\n"
```
bee-san, line 24: noice

pywhat/what.py
```diff
@@ -26,6 +26,13 @@ def main(text_input):
 
     Your text must either be in quotation marks, or use the POSIX standard of "--" to mean "anything after -- is textual input".
 
+
+    pyWhat can also search files or even the whole directory with recursion:
+
+        * what secret.txt
+
+        * what "this/is/a/path"
```
bee-san, line 34: We need single quotations here on both of them, not double `"` or no quotations. Google "Hashcat can't crack hash with `"`" to find out why Bash hates double quotation marks :-)

pywhat/identifier.py
amadejpapez: I have removed this, so if API was given a file path it will work the same. Or should it be done differently?

bee-san: No, it should work like that I believe! Make sure to include tests for if the API was given a file path :)

amadejpapez: Done!

tests/test_click.py
```diff
@@ -96,12 +96,12 @@ def test_file_fixture13():
...
     assert re.findall("Bitcoin", str(result.output))
 
-
+@pytest.mark.skip("Test does not work because of table overflow")
 def test_arg_parsing():
     runner = CliRunner()
     result = runner.invoke(main, ["1KFHE7w8BhaENAswwryaoccDb6qcT6DbYY"])
     assert result.exit_code == 0
-    assert re.findall("blockchain", str(result.output))
+    assert re.findall("blockch", str(result.output))
```
bee-san, line 104: Why does this test not work?
1. Tables auto-flow to the next line (if they do not, you need to edit the table to allow auto-flowing. See how I did it with one of the other columns)
2. Even if it did overflow, I'd assum `blockch` is in the test and thus the test passes?
3. This is why I love testing -- I get to see the system at play :D

tests/test_click.py
```diff
@@ -188,12 +188,12 @@ def test_file_fixture_youtube_id():
...
     assert re.findall("YouTube", str(result.output))
 
-
+@pytest.mark.skip("Test does not work because of table overflow")
 def test_file_fixture_ip4():
     runner = CliRunner()
     result = runner.invoke(main, ["fixtures/file"])
     assert result.exit_code == 0
-    assert re.findall("Address Version 4", str(result.output))
+    assert "4" in str(result.output)
```
bee-san, line 196: See above

pywhat/identifier.py
```diff
@@ -1,36 +1,48 @@
...
+            # if input is a directory, recursively search for all of the files
+            for myfile in glob.iglob(text + "**/**", recursive=True):
+                if os.path.isfile(myfile):
+                    get_file_data(myfile)
+
+        elif os.path.isfile(text):
+            get_file_data(text)
+
+        else:
+            file_text["None"] = [text]
```
bee-san, line 41: What does `None` here mean? Can we use a better variable name?

pywhat/identifier.py
```diff
@@ -1,36 +1,48 @@
...
+
+        def get_file_data(filepath):
+            short_name = os.path.basename(filepath)
+            magic_numbers[short_name] = self.file_sig.open_binary_scan_magic_nums(filepath)
+            file_text[short_name] = self.file_sig.open_file_loc(filepath)
+
+            # if file doesn't exist, check to see if the inputted text is
+            # a file in hex format
+            if not magic_numbers[short_name]:
+                magic_numbers[short_name] = self.file_sig.check_magic_nums(filepath)
```
bee-san, line 28: Functions in functions are not very pythonic, I suggest changing this :-) 

Functions should do one thing and only one thing. Try to follow that principle, if a function does multiple things it should be split into multiple functions. And the name of the function should be what it's doing.

Read this for more info:
https://www.amazon.co.uk/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882

pywhat/identifier.py
```diff
@@ -1,36 +1,48 @@
...
+
+
+        if os.path.isdir(text):
+            # if input is a directory, recursively search for all of the files
+            for myfile in glob.iglob(text + "**/**", recursive=True):
+                if os.path.isfile(myfile):
+                    get_file_data(myfile)
```
bee-san, line 35: This would be nicer in a loop with a variable assignment, it's hard to tell _where_ the data is going without reading the function :)

pywhat/identifier.py
```diff
@@ -1,36 +1,48 @@
...
+        elif os.path.isfile(text):
+            get_file_data(text)
+
+        else:
+            file_text["None"] = [text]
+
+        identify_obj["File Signatures"] = magic_numbers
+        identify_obj["Regexes"] = self.regex_id.check(file_text)
+        # get_hashes takes a list of hashes, we split to give it a list
+        # identify_obj["Hashes"] = self.name_that_hash.get_hashes(text.split())
```
bee-san, line 46: Delete this code unless it's needed?

pywhat/printer.py
```diff
@@ -26,38 +26,47 @@ def pretty_print(self, text: dict):
             table.add_column("Matched Text", overflow="fold")
             table.add_column("Identified as", overflow="fold")
             table.add_column("Description")
-            for i in text["Regexes"]:
-                matched = i["Matched"]
-                name = i["Regex Pattern"]["Name"]
-                description = None
+            table.add_column("Filename", overflow="fold")
```
bee-san, line 29: If your overflow is `fold`, why do your tests fail?

pywhat/printer.py
```diff
@@ -26,38 +26,47 @@ def pretty_print(self, text: dict):
...
+                            )
+                        else:
+                            description = i["Regex Pattern"]["Description"]
+
+                    if description:
+                        table.add_row(
+                            matched,
+                            name,
+                            description,
+                            filename
```
bee-san, line 58: Do we add a row on API input? Does that cause a problem?

pywhat/regex_identifier.py
```diff
@@ -12,42 +12,46 @@ def __init__(self):
...
-
-                    matches.append(
-                        {
-                            "Matched": matched,
-                            "Regex Pattern": reg,
-                        }
-                    )
+        matches = {}
+        for key, value in text.items():
+            for txt in value:
```
bee-san, line 17: Can we use better variables here? What is `txt`? Perhaps a type-hint for the function so I actually know what's in `text.items()`? 😄

tests/test_identifier.py
```diff
@@ -4,5 +4,14 @@
 def test_identifier_works():
     r = identifier.Identifier()
     out = r.identify("DANHz6EQVoWyZ9rER56DwTXHWUxfkv9k2o")
-    assert "Dogecoin (DOGE) Wallet Address" in out["Regexes"][0]["Regex Pattern"]["Name"]
+    assert "Dogecoin (DOGE) Wallet Address" in out["Regexes"]["None"][0]["Regex Pattern"]["Name"]
```
bee-san, line 7: Why is `"None"` added here? This looks like it'll break the API, everyone will have to update their own scripts -- no?

tests/test_regex_identifier.py
```diff
@@ -10,303 +10,303 @@ def test_regex_successfully_parses():
 
 def test_regex_runs():
     r = regex_identifier.RegexIdentifier()
-    res = r.check(["DANHz6EQVoWyZ9rER56DwTXHWUxfkv9k2o"])
-    assert "Dogecoin (DOGE) Wallet Address" in res[0]["Regex Pattern"]["Name"]
+    res = r.check({"test": ["DANHz6EQVoWyZ9rER56DwTXHWUxfkv9k2o"]})
```
bee-san, line 13: Why does every API input need to be in a dictionary? What are the values of this dictionary? What is `test`?

bee-san (dismissed):

tests/test_regex_identifier.py
amadejpapez: Sooo I firstly assigned every text to a file `{"filename": ["text"]}` and this part was then submitted to check(), so this is why I changed all the tests as it required a dictionary with a filename. I have already changed this and will push a commit with some of the other changes - now it assigns matches to a filename AFTER check() comes back with the results, so tests were reverted back. Did not notice this yesterday :')

tests/test_identifier.py
amadejpapez: This value is a filename, but if there is no file, I made to be None (that obviously was not the best idea) so that it could be easily printed in the table as None if there is no file. Will change this to "text" and when printing change it to "None".

pywhat/printer.py
amadejpapez: Yes it is also added with the API and it works the same without any problems - if there is a file, this is a filename; if not, it is a string `None`

tests/test_click.py
amadejpapez: These tests are really useful to see how everything works (and what does not). So I did some experimenting - I cloned main repo, added new `filename` column and same `test_click` tests started to fail.

This "blockchain" fails because only `blockc` is in one line. My latest commit changed this test to check for `blockc` and it works! As soon as `h` is added, it fails. So I would guess that test does not see over this line that is visible in the table.

image

tests/test_click.py
amadejpapez: Same as with `blockchain` test, this two IP address tests start to fail as soon as a new column is added and more text overflows into another line.

Previously it matched to `Address Version 4` and this is how the table looked:

image

Now that does not work as `Version` and `4` are each in their own line. So assertion in test does not see the text in the next line or just does not make the connection between the multiple lines. Maybe we could check for the string before printing in the table?

pywhat/identifier.py
amadejpapez: It is needed for the NTH part, so if there are no plans for that, NTH could be completely removed from pyWhat?

tests/test_click.py
bee-san: I am 99% sure the blockchain link is not clickable, this is because it cuts off the clickable part :-( 

I _think_ I didn't include the overflow originally because of that?

amadejpapez: That is correct yeah. It is not clickable and even with overflow I don't think the test would work as it still cuts text into another line which breaks clickable links and tests :(

bee-san: Can you show what a clickable link looks like?

amadejpapez: Sure! In my setup, I am unable to get or click on the full link that should be in the table. If I click directly or click "Open Link" in options, it will only get the part that is visible in the table. Everything after `...` is like it is not there.
image

Are you able to click and get the full link? Maybe this depends on the setup and OS but for me it is not working. Even if overflow was on, it would of course break clickable part and also break copying - as it selects the whole row, not only that one part.

Here I tried to copy the name:
image

I guess the issue is fundamentally in the the tables work (as it is seems it is structured by rows and not by columns) - probably because of the terminal, and I do not know what the solution for this could be.

amadejpapez: Maybe the solution could be printing all information in one row? Like this:
image
But this still brakes some longer URLs, so maybe a built in URL shortener or something? :laughing:

bee-san: So you're saying it's impossible to force Rich to display that URL? Can you open a github issue on their repo and ask them if theres a solution? :)

tests/test_regex_identifier.py
amadejpapez: Yes! I have changed this, so no changes are needed for the API and also tests were reverted back as they no longer need to be dictionaries :D

tests/test_click.py
amadejpapez: After your reply, I have went through some of the issues in rich repo and links over multiple lines cannot be made clickable in tables. When text/link is too long, rich inserts a new line and adds it to the next row, which causes links to break.

There are some solutions but they don't work for most terminals - [#866](https://github.com/willmcgugan/rich/issues/886#issuecomment-756406589) and [#428](https://github.com/willmcgugan/rich/issues/428#issuecomment-721143788). If we want links to work, we would have to change how the table looks. Maybe to have more information in one row - similar to my concept above and use a URL shortener so links are always in one line, or maybe to start using something else than tables altogether. Maybe you have any better solutions about what could be done about this? :/

pywhat/printer.py
amadejpapez: Nevermind. Only the `identifier` is executed in the API mode without `printer`, so there are no tables or rows in the API mode :)

But the API results are a little different as there is a `filename` added to each `Regexes` and `File Signatures`. It now also accepts file paths, so if input is a folder this will be quite useful. With normal text as input this is `text`.  An example:
```
{
    'File Signatures': {
        'text': None
        },

    'Regexes': {
        'text': [
            {
                'Matched': 'DANHz6EQVoWyZ9rER56DwTXHWUxfkv9k2o',
                'Regex Pattern': {
                    'Name': 'Dogecoin (DOGE) Wallet Address',
                    'Regex': '^D{1}[5-9A-HJ-NP-U]{1}[1-9A-HJ-NP-Za-km-z]{32}$',
                    'plural_name': False,
                    'Description': None,
                    'URL': 'https://dogechain.info/address/',
                    'Rarity': 0.5,
                    'Tags': [
                        'Cryptocurrency',
                        'Finance'
                        ]
                    }
                }
        ]
    }
}
```

tests/test_click.py
amadejpapez: So apparently clickable links work for some terminals, so this is not as big of an issue as I thought.

pywhat/identifier.py
```diff
@@ -1,36 +1,64 @@
...
+            text = self.file_sig.open_file_loc(input)
+
+            # if file doesn't exist, check to see if the inputted text is
+            # a file in hex format
+            if not magic_numbers:
+                magic_numbers = self.file_sig.check_magic_nums(text)
+
+            short_name = os.path.basename(input)
+            identify_obj["File Signatures"][short_name] = magic_numbers
+            identify_obj["Regexes"][short_name] = self.regex_id.check(text)
```
piatrashkakanstantinass, line 47: You wrote essentially the same code twice. Maybe you can shorten it? 😅

piatrashkakanstantinass (changes requested):

pywhat/identifier.py
amadejpapez: Well this was in a function before but I changed it after this two comments from Bee: [1](https://github.com/bee-san/pyWhat/pull/58#discussion_r643007439) and [2](https://github.com/bee-san/pyWhat/pull/58#discussion_r643008060)

piatrashkakanstantinass: Well, you can store all filenames in a list and iterate over it. This way you will not be repeating yourself twice.

amadejpapez: That's a good idea yeah. Will change it later. Thanks! :D

amadejpapez: I have changed this now and it looks much better 😄

piatrashkakanstantinass: Beautiful!

amadejpapez:
Okay @bee-san I think this PR is done now. I have reverted all of the changes to tests, so they all work as before. Also pyWhat is now searching filenames, so if there is a file found with recursion that has a CTF flag in the name, pyWhat will be able to identify it. :D

bee-san:
Conflicting merge issues :cry: Did you say that links didn't work on this PR? 🙏🏻

amadejpapez:
> Conflicting merge issues  Did you say that links didn't work on this PR? 🙏🏻

Just saw the new release yes, so will fix them! :grin:

About the links, this issue is affecting the latest version of pyWhat and not only this PR. Links only work for some terminals - I tested in Arch and Kali, and they do not open the full URL but only the _visible_ part. This PR is not causing this but when tests broke, I noticed that links do not open for me. Here is a screenrecording from Kali, running pyWhat from PyPi, so you can see what is happening:

https://user-images.githubusercontent.com/63184600/120925721-202e2f80-c6da-11eb-8ff8-3534397c9921.mp4

amadejpapez:
I have fixed almost every test, but as soon as I add recursion support, filtration brakes. I just cannot get my head around why.. Could you @piatrashkakanstantinass please take a look at this as you have written filtration? :pleading_face: 

If I run `pywhat --include_tags finance fixtures/file`, `distribution` in regex-identifier.py prints as `Distribution({'Tags': {'finance'}...` (what it should be) but in my version it is `Distribution({'Tags': {'password', 'credentials', 'hacking', 'geo-location', 'token', 'media', 'finance', 'internet', 'cryptocurrency', 'username', 'networking', 'website', 'security', 'aws', 'cyber security', 'identifiers', 'credit card'}...` so it is not seeing what to filter and it just returns all of the regexes.

I did not do any changes to the `distribution`, so something else needs to be changed.

bee-san:
> they do not open the full URL but only the visible part.
This is actually very normal and quite expected, hyperlinks are not supported in all terminals.

This is why it is essential that we show the **full link with no cut-off or new lines** to make it work everywhere :)

bee-san:
What if we removed:

> Click here to analyse in the browser

So there was more space? Or we changed the wording?

amadejpapez:
> So there was more space? Or we changed the wording?

This would sadly not fix it. The only way I was able to get it to work for me is if the URL is from the start to the end in one line, without `...`.

If we did something like this:
image

This makes links to work unless they again go over the line... 😢 Maybe we could add recursion to pyWhat and make this into an issue 😄

pywhat/regex_identifier.py
```diff
@@ -1,57 +1,58 @@
...
+                        codes_fullpath = os.path.join(
+                            os.path.dirname(os.path.abspath(__file__)), codes_path)
+                        with open(codes_fullpath, "r", encoding="utf-8") as myfile:
+                            codes = json.load(myfile)
+
+                        locations = []
+                        for code in codes:
+                            if number.startswith(code["dial_code"]):
+                                locations.append(code["name"])
+                        if len(locations):
```
SkeletalDemise, line 41: This isn't Pythonic. Change it to `if len(locations) > 0:`

pywhat/identifier.py
```diff
@@ -1,47 +1,64 @@
...
+        if distribution is None:
+            self.distribution = Distribution()
+        else:
+            self.distribution = distribution
+        self._regex_id = RegexIdentifier()
+        self._file_sig = FileSignatures()
+        self._name_that_hash = Nth()
+
+    def identify(self, input: str, dist: Distribution = None,
+                 api=False) -> dict:
```
SkeletalDemise, line 22: Why did you get rid of all the `api` stuff? With your changes it breaks the pywhat checker in Ciphey. Ciphey calls pywhat with `self.id.identify(ctext, api=True)`

pywhat/identifier.py
```diff
@@ -1,47 +1,64 @@
...
-        identify_obj["Regexes"] = self._regex_id.check(text, dist)
-
-        # get_hashes takes a list of hashes, we split to give it a list
-        # identify_obj["Hashes"] = self._name_that_hash.get_hashes(text.split())
-
-        return identify_obj
-
-    def _file_exists(self, text):
-        return os.path.isfile(text)
+import os.path
```
SkeletalDemise, line 1: Run `isort` on the files to sort the imports please.

SkeletalDemise (changes requested):

pywhat/regex_identifier.py
amadejpapez: This was there before, but I have changed it now :)

pywhat/identifier.py
amadejpapez: I looked into Ciphey and the part that is breaking it is that now `Regexes` are inside of a dictionary that is holding a file name - as API also [supports](https://github.com/bee-san/pyWhat/pull/58#discussion_r642871675) files and folders now. If there is no file, this value is just `text`. For it to work you would have it change this line:

```
matched_regex = returned_regexes["Regexes"][0]["Regex Pattern"]
To this:
matched_regex = returned_regexes["Regexes"]["text"][0]["Regex Pattern"]
```
But this will cause errors to people who already installed Ciphey right? Hmmm maybe API should not support files and folders?

pywhat/identifier.py
amadejpapez: Nice! I didn't know about this :)

pywhat/identifier.py
SkeletalDemise: That's fine. I'll just change it in Ciphey and bump up the pywhat version in `pyproject.toml`

amadejpapez: Oh okay. Amazing! :D

SkeletalDemise (approved):

Issue #29, opened 2021-05-22
Add filtration to API and CLI


bee-san:
Should be able to filter by rarity and tags :-)

bee-san:
@piatrashkakanstantinass is assigned to this issue

bee-san:
Fixed in #62

PR #62, opened 2021-05-31
Filtration support

piatrashkakanstantinass:
**Add filtration support(#29)**

- You can check available tags by using `pywhat --tags` or
```
from pywhat import *
print(pywhat_tags)
```
- Filtration CLI:
`pywhat --rarity min:max --include_tags tag1,tag2 --exclude_tags tag1,tag2 TEXT`
'min' and 'max' can be omitted.

**ToDo:**

- [x] Finish work on distributions
- [x] Update CLI
- [x] Move API to \_\_init\_\_.py' for easy access
- [x] TESTS
- [x] Clean up the mess


bee-san:
Ahh!!! This is gonna be annoying, but I am actually working on a filter system myself :-) See my branch here: https://github.com/bee-san/pyWhat/tree/bee-filter

However!! I really like some of the things here. The arguments, listing tags, and docs are great!!

Here's how I wanted to do it:
**Problem**: We have a program which requires 2 filters, this is a real-world example. [Ciphey](https://github.com/ciphey/ciphey) would like to have a filter for checking whether the item is possible plaintext (bitcoin address, email, website) but also a filter for checking for _encoded_ text, like JWT tokens or Morse code etc etc.

So, we have 1 application which requires 2 filters:
1. Plaintext identification 
2. Stuff that needs to be decoded, but is the answer

By implementing filtering at the identifier level, every time ciphey runs it'd have to re-create the filters with the API. This means that if it runs 10 times a second, it re-makes the filters 10 times a second.

You can get around this by filtering at the object attributes level, so you make the object once and the filters are there -- but this prevents you from changing the filters or adding new ones.

You can of course cheat and manually make the object, but that's cheating and I'd prefer a high level API for users.

I propose adding _distributions_. A _distribution_ is a regex list which has been filtered (see my branch https://github.com/bee-san/pyWhat/tree/bee-filter/pywhat/filtration_distribution ).

Then, the regex_identifier (or identifier) takes this distribution object and uses the `distribution.get_regexes()` to automatically get the already tagged data.

This means we can:
1. Have as many filtration systems as we want without re-making them everytime.
2. Be nimble enough that we can have more than 1 filtration system.

I also envisioned adding magic methods to distributions like `__add__`, so you can add 2 distributions together or take away one distribution from another. 

So going back to the Ciphey example, I imagine something like this:
1. Plaintext identification

```py
filters = {"Tags": ["Networking", "Credentials"], "Min_Rarity": 0.6}
distribution = Distribution(filters)
```

And then adding decoding stuff:

```py
filters1 = {"Tags": ["Needs Decoding", "IPv6]} # This tag doesn't exist yet, we also want short names for each regex as a tag in the future
distribution2 = Distribution(filters1)
```

In total:

```py
filters = {"Tags": ["Networking", "Credentials"], "Min_Rarity": 0.6}
distribution = Distribution(filters)

filters1 = {"Tags": ["Needs Decoding", "IPv6]} # This tag doesn't exist yet, we also want short names for each regex as a tag in the future
distribution2 = Distribution(filters1)

identifier("text here", distribution)
```

So we pass the identifier the distribution object.

**If no distribution object is passed, we should make a distribution object at the start of the program which is everything with no filters**

Hope that makes sense!! Sorry I haven't had much time to do this, I've been busy 😅

If you fancy picking up my branch you can always merge what I have with yours? 🥺🙏

Thank you **so** much for contributing!!!

piatrashkakanstantinass:
Sounds interesting!

bee-san:
> Sounds interesting!

Are you interested in picking this up or should I finish it? 😄

piatrashkakanstantinass:
> > Sounds interesting!
> 
> Are you interested in picking this up or should I finish it? 😄

I will finish it. However, need some time😀

piatrashkakanstantinass:
I am thinking about making API look like this:
```
...
distribution1 = Distribution(filter1)
distribution2 = Distribution(filter2)
id = identifier.Identifier(distribution1)
id.identify(text)
id.distribution = distribution2
id.identify(text)
```

What are your thoughts, @bee-san?

piatrashkakanstantinass:
**ToDo:**

- [ ] Finish work on distributions
- [ ] Update CLI
- [ ] Move API to __init__.py for easy access

bee-san:
> I am thinking about making API look like this:
> 
> ```
> ...
> distribution1 = Distribution(filter1)
> distribution2 = Distribution(filter2)
> id = identifier.Identifier(distribution1)
> id.identify(text)
> id.distribution = distribution2
> id.identify(text)
> ```
> 
> What are your thoughts, @bee-san?

I can see where you're going, but also I think repeatedly switching the variables over is a bad idea? In an ideal world the API would look like:

```py
distribution1 = Distribution(filter1)
distribution2 = Distribution(filter2)
id = identifier.Identifier() # No distribution

id.identify(text, distribution1)
id.identify(text, distribution2)
id.identify(text) # Uses no distribution, which means it uses "everything". 
```

You can achieve that last one by creating a distribution for the whole program with everything,. and setting it as the default in the function like:

```py
class Identifier():
    def __init__(self):
        self.default_distribution = distribution()
    def identify(self, distribution=self.default_distribution)
```

That way it either uses:
1. The distribution passed to it
2. The default one

I just dislike the idea of repeatedly changing a variable in an object, I'd much rather it be more _functional_ like 😄 

Thanks so much for the ❓ !

bee-san:
@piatrashkakanstantinass 

>  Move API to __init__.py' for easy access

Can you explain how this works? I'm not familiar with this pattern 😄

piatrashkakanstantinass:
> @piatrashkakanstantinass
> 
> > Move API to **init**.py' for easy access
> 
> Can you explain how this works? I'm not familiar with this pattern 😄

After applying the changes, API will be like this:
```
import pywhat
id = pywhat.Identifier()
dist = pywhat.Distribution(some_filter)
id.identify(some_text, dist)
```
Importing `pywhat` and getting all the functionality is more straightforward than importing `pywhat.identifier`

piatrashkakanstantinass:
@bee-san, I guess supporting distribution both as an attribute of `Identifier` and as an optional parameter for `identify` will leave everyone happy?🤔

bee-san:
> @bee-san, I guess supporting distribution both as an attribute of `Identifier` and as an optional parameter to `identify` will make everyone happy?🤔

Sure! That makes sense :) I can see why having multiple-identifiers might be handy if the program has hundreds of filters 😄  But also I see why having one identifier is cool if it only has one or two :)

piatrashkakanstantinass:
Now I am really confused on what went wrong.

piatrashkakanstantinass:
Soo, now it is possible to import API by using `from pywhat import *`
### Distributions
```python
from pywhat import *
dist = Distribution({"Tags": ["Identifiers"], "ExcludeTags": ["Credentials"], "MinRarity": 0.2, "MaxRarity": 0.8})
id = Identifier()
res = id.identify(DATA, distribution=dist)

dist2 = dist & Distribution({"Tags": ["Finance", "Media"]}) # not supported yet :(
id.distribution = dist2
id.identify(DATA)
```

Tests are urgently important and it would be great if someone could help me with that 😏
I am going to add dunder methods to Distributions for set-like behaviour. Probably implementing some helper class for work with filters_dict's is something I should do, too. Also Distribution class must strictly check filters_dict's since it is really simple to forget brackets (like this: `{"Tags": "some_tag"}`).

P.S.: I really dislike how CLI option parsing looks like, I should move parsing to another function or check if click has some advanced solutions to offer.

bee-san:
Sure! I'll perhaps work on some tests later, I have been very sick this week but I am feeling better 😄

bee-san:
@piatrashkakanstantinass are you in the discord? http://discord.skerritt.blog :)

piatrashkakanstantinass:
> @piatrashkakanstantinass are you in the discord? http://discord.skerritt.blog :)

Sure! r49behind#6377 :)

piatrashkakanstantinass:
So, I have written some tests for distributions. Tests for identifiers still need to be written. Also, I decided not to create a Filter helper class since it does not offer any significant functionality.

bee-san:
hii!!! nice!!! do we not already have identifier tests? Or do you mean identifiers with Filters:tm: tests? :)

piatrashkakanstantinass:
> hii!!! nice!!! do we not already have identifier tests? Or do you mean identifiers with Filters™️ tests? :)

Yes, we do, but I want to test the actual behaviour of API with Filters™️XD

piatrashkakanstantinass:
@bee-san Well, it is done, I guess...

bee-san:
Let me review 😄 !!!!!

pywhat/distribution.py
```diff
@@ -0,0 +1,95 @@
...
+    Example filters:
+    * {"Tags": ["Networking"]}
+    * {"Tags": ["Identifiers"], "ExcludeTags": ["Credentials"], "MinRarity": 0.6}
+    """
+
+    def __init__(self, filters_dict: Optional[dict] = None):
+        tags = AvailableTags().get_tags()
+        self._dict = dict()
+        if filters_dict is None:
+            filters_dict = dict()
```
bee-san, line 22: We don't need this because of the function argument, no? We can change it:

```diff
- def __init__(self, filters_dict: Optional[dict] = None):
+ def __init__(self, filters_dict: Optional[dict] = {}):
```

pywhat/distribution.py
```diff
@@ -0,0 +1,95 @@
...
+
+        self._dict["Tags"] = set(filters_dict.setdefault("Tags", tags))
+        self._dict["ExcludeTags"] = set(filters_dict.setdefault("ExcludeTags", set()))
+        self._dict["MinRarity"] = filters_dict.setdefault("MinRarity", 0)
+        self._dict["MaxRarity"] = filters_dict.setdefault("MaxRarity", 1)
```
bee-san, line 27: This is cool!

pywhat/distribution.py
```diff
@@ -0,0 +1,95 @@
...
+    def get_regexes(self):
+        return list(self._regexes)
+
+    def get_filter(self):
+        return dict(self._dict)
+
+    def __repr__(self):
+        return f"Distribution({self._dict})"
+
+    def __and__(self, other):
```
bee-san, line 64: I would have gone with `__add__` here,  but I suppose it makes more natural sense to use `__and__`

pywhat/distribution.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def __iand__(self, other):
+        if type(self) != type(other):
+            return NotImplemented
+        return self & other
+
+    def __ior__(self, other):
+        if type(self) != type(other):
+            return NotImplemented
+        return self | other
```
bee-san, line 95: These are cool, I wasn't aware of this! :)

pywhat/helper.py
```diff
@@ -0,0 +1,25 @@
...
+import os.path
+
+
+class AvailableTags():
+    def __init__(self):
+        self.tags = set()
+        path = "Data/regex.json"
+        fullpath = os.path.join(os.path.dirname(os.path.abspath(__file__)), path)
+        with open(fullpath, "r", encoding="utf-8") as myfile:
+            regexes = json.load(myfile)
```
bee-san, line 12: We load regexes in ```    def _load_regexes(self):``` already, maybe abstract that and make that function exist only in the helper file.

Then just call that function in `distribution.py`?

pywhat/what.py
```diff
@@ -1,5 +1,47 @@
...
+            print("Invalid rarity range format ('min:max' expected)")
+            sys.exit(1)
+        try:
+            if not rarities[0].isspace() and rarities[0]:
+                filter["MinRarity"] = float(rarities[0])
+            if not rarities[1].isspace() and rarities[1]:
+                filter["MaxRarity"] = float(rarities[1])
+        except ValueError:
+            print("Invalid rarity argument (float expected)")
+            sys.exit(1)
```
bee-san, line 33: Nice parsing here!

tests/test_regex_identifier.py
```diff
@@ -4,8 +4,8 @@
 
 def test_regex_successfully_parses():
     r = regex_identifier.RegexIdentifier()
-    print(r.regexes)
-    assert "Name" in r.regexes[0]
+    print(r.distribution.get_regexes)
```
bee-san, line 7: Ahh delete this if you can! I messed up here 😢

bee-san (commented):
Looks great!!!! I'll write some docs based on your tests 😄  Thanks so much! Just a few comments and questions :-)

pywhat/distribution.py
piatrashkakanstantinass: Yes, we can; however, I don't think it would be great since if somebody ever decides to change the code and mutate filters_dict, python will produce unexpected behaviour. So, in my opinion, it is better to write 2 extra lines of code rather than to be confused by the way python behaves if you ever decide to change the code. What do you think?

pywhat/distribution.py
piatrashkakanstantinass: Yeah, I have been thinking about __add__, but its behaviour with rarities would be unclear, so I used bitwise operators instead.

bee-san:
Small update! I checked out this branch, will test manually and update docs then accept :)

tests/test_click.py
```diff
@@ -11,6 +13,22 @@ def test_hello_world():
     assert "THM{" in result.output
 
 
+def test_filtration():
+    runner = CliRunner()
+    result = runner.invoke(main, ["--rarity", "0.5:", "--include_tags", "Identifiers,Media", "fixtures/file"])
```
bee-san, line 18: This test doesn't actually filter for me. 

```
python3 -m pywhat --rarity 0.1:0.2 --include_tags "Identifiers,Media" "fixtures/file"
```
image

tests/test_click.py
```diff
@@ -11,6 +13,22 @@ def test_hello_world():
...
+    assert "THM{" in result.output
+    assert "ETH" in result.output
+
+
+def test_tag_printing():
+    runner = CliRunner()
+    result = runner.invoke(main, "--tags")
+    assert result.exit_code == 0
+    for tag in pywhat_tags:
+        assert tag in result.output
```
bee-san, line 29: Do you think tags should be non-case sensitive? 
image

tests/test_click.py
bee-san: Tags do not also work from the CLI:
image

bee-san: Also, what's up with the double `file identified`?
image

tests/test_identifier.py
```diff
@@ -6,3 +7,21 @@ def test_identifier_works():
     out = r.identify("DANHz6EQVoWyZ9rER56DwTXHWUxfkv9k2o")
     assert "Dogecoin (DOGE) Wallet Address" in out["Regexes"][0]["Regex Pattern"]["Name"]
 
+
+def test_identifier_filtration():
+    filter = {"Tags": ["Password"]}
+    r = identifier.Identifier(Distribution(filter))
+    regexes = r.identify('fixtures/file')["Regexes"]
+    for regex in regexes:
+        assert "Password" in regex["Regex Pattern"]["Tags"]
```
bee-san, line 16: This is super weird because the API works fine but the CLI tool doesn't! :(

pywhat/what.py
```diff
@@ -8,14 +50,26 @@
     )
 )
 @click.argument("text_input", required=True)
-def main(text_input):
+@click.option("--tags", is_flag=True, expose_value=False, callback=print_tags, help="Show available tags and exit.")
+@click.option("--rarity", help="Filter by rarity.")
+@click.option("--include_tags", help="Only print entries with included tags.")
+@click.option("--exclude_tags", help="Exclude tags.")
```
bee-san, line 56: Can we assign short names here? Just copy what I did (I don't wanna write to your branch and potentially over-write stuff you've done)

```py
@click.argument("text_input", required=True)
@click.option("-t", "--tags", is_flag=True, expose_value=False, callback=print_tags, help="Show available tags and exit.")
@click.option("-r", "--rarity", help="Filter by rarity. This is in the range of 0:1. To filter only items past 0.5, use 0.5: with the colon on the end.")
@click.option("-i", "--include_tags", help="Only print entries with included tags.")
@click.option("-e", "--exclude_tags", help="Exclude tags.")
```

tests/test_click.py
bee-san: Perhaps we should also write some tests to make sure this doesn't happen again, I.E. if `--rarity 0:0` or `--rarity -1:-1`  there should be no output? 
image

bee-san (changes requested):
Some bugs with the CLI tool itself, but the API strangely works fine! Will write docs for the API :)

piatrashkakanstantinass:
Okay, going to work on that😀

piatrashkakanstantinass:
Aww, I forgot to return distribution in parse_options() 😅

tests/test_click.py
piatrashkakanstantinass: > What's up with the double `file identified`?
> image

IDK, it is present in the main branch, too

bee-san:
> Aww, I forgot to return distribution in parse_options() 😅

Ahh always the way! The great news is that after this (and after I write some docs 😉 ) we can release!!!

bee-san:
On it! ✍🏻

bee-san:
It works!!!
image

bee-san:
Docs:
https://github.com/bee-san/pyWhat/wiki/API

bee-san (approved):

Issue #70, opened 2021-06-07
Add sorting support


piatrashkakanstantinass:
Since we have filtration, it is logical to assume that sorting should be supported, too.
Therefore, I propose the following changes to API:
```python
from pywhat import *
id = Identifier()
out = id.identify('fixtures/file', key=keys.rarity, reversed=True) # out[0] has the highest rarity
out = id.identify('fixtures/file', key=keys.matched) # sort alphabetically by matched string
out = id.identify('fixtures/file', key=keys.name) # sort by regex name
```
Also, as with distributions, it would be great to make it possible to specify keys while instantiating Identifier.
Something like that:
```python
id = Identifier(key=keys.name)
id.identify('fixtures/file') # identical to id.identify('fixtures/file', key=keys.name)
```

PR #71, opened 2021-06-08
Add sorting

piatrashkakanstantinass:
#70
- Now, all parameters to identify (except text input) and Identifier are keyword-only.
- Added sorting
Example:
```python
from pywhat import *
id = Identifier()
id.identify(text, key=Keys.NAME, reverse=True) # sort by the name of regex pattern in reversed order
id2 = Identifier(key=Keys.MATCHED)
id2.identify(text) # sort by matched string
```

Available keys:
```python
Keys.NAME # Sort by the name of regex pattern
Keys.RARITY # Sort by rarity
Keys.MATCHED # Sort by matched string
Keys.NONE # No sorting is done (the default)
```

I am not sure how sorting should be implemented in terms of CLI.


pywhat/identifier.py
```diff
@@ -1,26 +1,46 @@
...
 from pywhat.nameThatHash import Nth
 from pywhat.regex_identifier import RegexIdentifier
 
 
 class Identifier:
-    def __init__(self, distribution: Optional[Distribution] = None):
-        if distribution is None:
+    def __init__(
+        self,
+        *,
```
bee-san, line 14: What's `*` here?

pywhat/identifier.py
```diff
@@ -36,7 +56,9 @@ def identify(self, text: str, dist: Distribution = None,
             # a file in hex format
             identify_obj["File Signatures"] = self._file_sig.check_magic_nums(text)
 
-        identify_obj["Regexes"] = self._regex_id.check(text, dist)
+        identify_obj["Regexes"] = sorted(
+            self._regex_id.check(text, dist), key=key, reverse=reverse
+        )
```
bee-san, line 61: Does this run everytime. even when no `key` is provided and it defaults? I was actively trying to take advantage of the fact that the json file is sorted and you don't need to sort after-the-fact as this saves a lot of time when repeatedly calling the API 😄

tests/test_identifier.py
```diff
@@ -1,27 +1,90 @@
...
         assert "Password" in regex["Regex Pattern"]["Tags"]
 
-    
+
 def test_identifier_filtration2():
     filter1 = {"ExcludeTags": ["Identifiers"]}
     filter2 = {"Tags": ["Identifiers"], "MinRarity": 0.6}
-    r = identifier.Identifier(Distribution(filter1))
-    regexes = r.identify('fixtures/file', dist=Distribution(filter2))["Regexes"]
+    r = identifier.Identifier(dist=Distribution(filter1))
```
bee-san, line 26: API change? 🤔

bee-san (changes requested):

pywhat/identifier.py
piatrashkakanstantinass: Arguments following it are forced to be keyword-only.` identify()` already accepts a lot of arguments that do not have any priority, and their order may change in the future, so, in my opinion, it is better to force a user not to use positional arguments. Identifier's` __init__` accepts quite a lot of arguments, too.

tests/test_identifier.py
piatrashkakanstantinass: Yes😢

pywhat/identifier.py
piatrashkakanstantinass: Yes, it does. I may add keys.none key that would not do any sorting.

piatrashkakanstantinass:
So, now keys are uppercase (Keys.MATCHED, Keys.NONE, etc...)
I have added Keys.NONE key. That is the default key used by Identifier when user does not pass any keys. Keys.NONE disables sorting.

pywhat/identifier.py
bee-san: Fair enough! :)

pywhat/identifier.py
bee-san: Please! 😄  Just a small scale thing

tests/test_identifier.py
bee-san: :cry: its okay we'll make this 2.0.0 :)

pywhat/identifier.py
bee-san: Hey, not sure how:
>  Keys.NONE disables sorting.
Disables sorting if it still runs `sorted()`? 

Other than that, everything looks fine! :D

pywhat/identifier.py
```diff
@@ -1,61 +1,87 @@
...
+
+        for key_, value in identify_obj.items():
+            # if there are zero regex or file signature matches, set it to None
+            if len(identify_obj[key_]) == 0:
+                identify_obj[key_] = None
+
+        if key != Keys.NONE:
+            identify_obj["Regexes"][short_name] = sorted(
+                identify_obj["Regexes"][short_name], key=key, reverse=reverse
+            )
```
bee-san, line 85: nvm i just read this its perfect!

bee-san (approved):

bee-san:
Make sure to edit the API docs, please! :-)

bee-san (approved):

PR #72, opened 2021-06-09
Recursion fixes

amadejpapez:
- [x] `Filename` name in the column is now actually there
- [x] `Filename` column is only printed if input is a folder
- [ ] Fix links
- [ ] `::` should not match as IPv6
- [ ] add quotes to example in `what.py`, line 83
- [ ] make JWT regex require one letter, so it no longer matches to IPv4 addresses

### DIFF
--- a/pywhat/printer.py
+++ b/pywhat/printer.py
@@ -1,11 +1,12 @@
 import json
+import os
 
 from rich.console import Console
 from rich.table import Table
 
 
 class Printing:
-    def pretty_print(self, text: dict):
+    def pretty_print(self, text: dict, text_input):
         console = Console(highlight=False)
 
         to_out = ""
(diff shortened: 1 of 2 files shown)


bee-san:
Do you want to fix the jwt issue? You can probably add a letter for now but make a test which fails, and then skip it so we can fix it in the future?

amadejpapez:
> Do you want to fix the jwt issue? You can probably add a letter for now but make a test which fails, and then skip it so we can fix it in the future?

That's a good idea yeah. Will do that :)

You also mentioned that there aren't quotes on every example given in the help menu. On line 83, email should be in quotes right?

bee-san:
Line 83 of what?

bee-san:
yes it should :)

bee-san:
I'm going to merge this because I want much smaller PRs. You should make one PR per fix, for example:

>  Fix links

 this is incredibly hard and will delay your entire PR

>  add quotes to example in what.py, line 83

This is incredibly easy and can be merged quite quickly

Please stick to the [Agile Method](https://www.cprime.com/resources/what-is-agile-what-is-scrum/#:~:text=Agile%20software%20development%20refers%20to,%2Dorganizing%20cross%2Dfunctional%20teams.&text=Scrum%20and%20Kanban%20are%20two%20of%20the%20most%20widely%20used%20Agile%20methodologies.) and make many smaller PRs :D

PR #73, opened 2021-06-10
Detection for credit card numbers with spaces and more regex

amadejpapez:
- improved MasterCard, Visa, Discover, Maestro, Diners and American Express regex so numbers with spaces in between like `5555 5555 5555 4444` also get matched. Note that spaces need to be in the right place. Some spaces can be missing but they cannot be in the wrong place as the regex is looking: 4 numbers, is there a space?, 4 numbers, and so on. So `555555555555 4444` will get matched but something like `555 5 5555 5555 4444` would not. This can get changed but currently only the correct space format is supported. (added a test for each format with spaces)
- JWT Token regex now requires one number and one letter. They look like this: `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWV9.TJVA95OrM7E2cBab30RMHrHDcEfxjoYZgeFONFh7HgQ`, so something like an IPv4 address no longer matches to it
- IPv6 regex now requires at least 1 number. I did not want to change the regex too much to not break actual IPv6 detection, but a restriction like this stops it from matching to `::`, `::a`, `::]`
- fixed boundaries as some of the other regexes with different formats in them would also match inside of strings (Solo Card Number, Switch Card Number and Visa)

This is my regex PR. I also added quotes to email in `what.py`, so that should all be fixed now. Will be working on fixing links now :)


bee-san:
> IPv6 regex now requires at least 2 numbers. I did not want to change the regex too much to not break actual IPv6 detection, but a restriction like this stops it from matching to ::, ::7, ::a, ::5, ::]

My friend.... `::1: is a valid IPv6 address, it is equivalent to `127.0.0.1` 😜 

This is why I want much smaller PRs, because now you have to fix it! :(

pywhat/Data/regex.json
```diff
@@ -212,7 +212,7 @@
     },
     {
       "Name": "JSON Web Token (JWT)",
-      "Regex": "^([a-zA-Z0-9_=]+)\\.([a-zA-Z0-9_=]+)\\.([a-zA-Z0-9_\\-\\+\/=]*)",
+      "Regex": "(?i)^(?=.*[a-z])(?=.*[0-9])(?:[a-z0-9_=]+\\.){2}(?:[a-z0-9_\\-\\+\/=]*)$",
```
bee-san, line 215: If this was YAML we could totally write a comment here explaining this.... >:(

amadejpapez: I can break it down here! :smile:  So `(?i)` means case-insensitive (that is why I removed A-Z). `(?=.*[a-z])` means look-ahead - basically it is checking if in the whole match there is at least one letter from a-z. Then there is this `(?:[a-z0-9_=]+\\.)` which is the first part of the JWT. As first and second parts are the same, I added them together so there is now `(?:[a-z0-9_=]+\\.){2}` - two times. And then the last part. I also added `$` at the end as it was missing, so boundary was not working here - that is the main reason why it matched to a lot more than just JWT. If first few characters seemed like a JWT, it would match :D

amadejpapez:
> My friend.... `::1` is a valid IPv6 address, it is equivalent to `127.0.0.1` 😜 

Oh😢 will lower it to only one number

> This is why I want much smaller PRs, because now you have to fix it! :( 

Well I thought only regex changes would be okay for one PR :( Will fix this and then it is done!

Btw I have also updated the API docs. I updated the example and added explained recursion and searching files :)

pywhat/Data/regex.json
amadejpapez: Ohhh that is unfortunate that you cannot add comments yeah :/

pywhat/what.py
```diff
@@ -80,7 +80,7 @@ def main(text_input, rarity, include_tags, exclude_tags):
 
         * what -- 52.6169586, -1.9779857
```
bee-san, line 81: quotes here too

amadejpapez: I thought about that but didn't add it because of this line: `Your text must either be in quotation marks, or use the POSIX standard of "--" to mean "anything after -- is textual input".`

Issue #45, opened 2021-05-27
Better URL Regex


bee-san:
We do not match on:
```
google.com
```

Because there is no HTTPS at the start.

The solution is to build a list of all top level domains ( see https://data.iana.org/TLD/tlds-alpha-by-domain.txt for all of them in a neat text file) and match only if the end matches a TLD.

An example regex is:

```
.*\.com|\.org
```

Which matches:
```
tryhackme.com
```

PR #74, opened 2021-06-11
Improved detection of URLs with a valid list of TLDs

amadejpapez:
- a script for getting a list of valid TLDs was already added, this PR only adds the list to the URL regex
- URL regex no longer requires `http`, `https` or `ftp` at the start (https://github.com/bee-san/pyWhat/issues/45), except if it is an IP address
- added a few new tests, removed skip from `tryhackme.com` as it now matches and added a skip to `http://папироска.рф` as it no longer matches. Browser shows it as that but it understands it in a punycode form `https://xn--80aaxitdbjk.xn--p1ai/` - which the regex matches
- removed `langdetect` from `pyproject.toml`
- also reran `poetry install`, so `poetry.lock` is up to date


amadejpapez:
I need to do some more regex testing, so this is not finished yet. :)

bee-san:
Want to review? There's a merge conflict too but it's probably just black :)

amadejpapez:
Should be good now! :)

pywhat/Data/tld_list.txt
```diff
@@ -0,0 +1 @@
...--IO0A7I|XN--J1AEF|XN--J1AMH|XN--J6W193G|XN--JLQ480N2RG|XN--JLQ61U9W7B|XN--JVR189M|XN--KCRX77D1X4A|XN--KPRW13D|XN--KPRY57D|XN--KPUT3I|XN--L1ACC|XN--LGBBAT1AD8J|XN--MGB9AWBF|XN--MGBA3A3EJT|XN--MGBA3A4F16A|XN--MGBA7C0BBN0A|XN--MGBAAKC7DVF|XN--MGBAAM7A8H|XN--MGBAB2BD|XN--MGBAH1A3HJKRD|XN--MGBAI9AZGQP6J|XN--MGBAYH7GPA|XN--MGBBH1A|XN--MGBBH1A71E|XN--MGBC0A9AZCG|XN--MGBCA7DZDO|XN--MGBCPQ6GPA1A|XN--MGBERP4A5D4AR|XN--MGBGU82A|XN--MGBI4ECEXP|XN--MGBPL2FH|XN--MGBT3DHD|XN--MGBTX2B|XN--MGBX4CD0AB|XN--MIX891F|XN--MK1BU44C|XN--MXTQ1M|XN--NGBC5AZD|XN--NGBE9E0A|XN--NGBRX|XN--NODE|XN--NQV7F|XN--NQV7FS00EMA|XN--NYQY26A|XN--O3CW4H|XN--OGBPF8FL|XN--OTU796D|XN--P1ACF|XN--P1AI|XN--PGBS0DH|XN--PSSY2U|XN--Q7CE6A|XN--Q9JYB4C|XN--QCKA1PMC|XN--QXA6A|XN--QXAM|XN--RHQV96G|XN--ROVU88B|XN--RVC1E0AM3E|XN--S9BRJ9C|XN--SES554G|XN--T60B56A|XN--TCKWE|XN--TIQ49XQYJ|XN--UNUP4Y|XN--VERMGENSBERATER-CTB|XN--VERMGENSBERATUNG-PWB|XN--VHQUV|XN--VUQ861B|XN--W4R85EL8FHU5DNRA|XN--W4RS40L|XN--WGBH1C|XN--WGBL6A|XN--XHQ521B|XN--XKC2AL3HYE2A|XN--XKC2DL3A5EE0H|XN--Y9A3AQ|XN--YFRO4I67O|XN--YGBI2AMMX|XN--ZFR164B|XXX|XYZ|YACHTS|YAHOO|YAMAXUN|YANDEX|YE|YODOBASHI|YOGA|YOKOHAMA|YOU|YOUTUBE|YT|YUN|ZA|ZAPPOS|ZARA|ZERO|ZIP|ZM|ZONE|ZUERICH|ZW
```
bee-san, line 1: add this file to .gitignore and delete it from the repo! :)

pywhat/regex_identifier.py
```diff
@@ -16,8 +16,15 @@ def check(self, text, distribution: Optional[Distribution] = None):
             distribution = self.distribution
         matches = []
 
+        with open("pywhat/Data/tld_list.txt", "r") as file:
+            tlds = file.read()
```
bee-san, line 20: Can we make the python script do this, so we don't do it everytime we load it? I'd rather it all be in one regex in the fiile to start with :-)

bee-san (changes requested):
good work!!!

pywhat/regex_identifier.py
amadejpapez: Ooo that's a good idea!!

amadejpapez: I have changed this now. The script is working but the regex is hardcoded there, so if we change it in the database it needs to be changed in the script too. If I can do this somehow differently, let me know :sweat_smile:

bee-san: nah nah hard-coding URL regex is how i would do it too :-)

scripts/get_tlds.py
```diff
@@ -1,8 +1,22 @@
...
+    database = file.read()
+
+database = re.sub(
+    # what you want to replace
+    r'("Name": "Uniform Resource Locator \(URL\)",\s*"Regex": ")[^"]+(")',
+
+    # what you want to replace it with - need to be double slashes
+    r"\1(?i)^(?:(?:https?|ftp):\/\/(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(?::[0-9]{1,5})?)([a-z0-9-_\/]?\\\\.?)*$|^(?:(?:https?|ftp):\/\/)?(?:\\\\S+:\\\\S+@)?(?:[a-z0-9-_]?\\\\.?)+?(?:[a-z0-9-_][a-z0-9-_]{0,62})\\\\.(?:" + final_string + r")(?::\\\\d{2,5})?(?:[\/?#]\\\\S*)?$\2",
+
+    database
```
bee-san, line 18: This is very hard-core I would have just:
* Turned the file into JSON
* Find the URL regex's json
* Update the URL regex in the python dictionary
* `json.dump()` the dictionary back into the file

I don't really trust regex well enough to do this, but if it works it works!

tests/test_regex_identifier.py
```diff
@@ -93,12 +105,19 @@ def test_ip4():
...
     res = r.check(["http://папироска.рф"])
     assert "Uniform Resource Locator (URL)" in res[0]["Regex Pattern"]["Name"]
 
 
+def test_same_international_url_in_punycode():
+    r = regex_identifier.RegexIdentifier()
+    res = r.check(["https://xn--80aaxitdbjk.xn--p1ai/"])
+    assert "Uniform Resource Locator (URL)" in res[0]["Regex Pattern"]["Name"]
+
+
```
bee-san, line 120: Nice test!

bee-san (approved):

scripts/get_tlds.py
bee-san: Occams Razor:

> a scientific and philosophical rule that entities should not be multiplied unnecessarily which is interpreted as requiring that the simplest of competing theories be preferred to the more complex or that explanations of unknown phenomena be sought first in terms of known quantities.

AKA:

> The simplest solutions are always the best

And:

> The golden hammer, if you are familiar with a hammer then everything looks like a nail

That doesn't mean everything is best solved with a hammer, you're using regex here because you're most familiar with it I believe. Sometimes we have to step back and look at better tooling 😄 
https://en.wikipedia.org/wiki/Law_of_the_instrument

amadejpapez: > This is very hard-core I would have just:
> 
> * Turned the file into JSON
> 
> * Find the URL regex's json
> 
> * Update the URL regex in the python dictionary
> 
> * `json.dump()` the dictionary back into the file
> 
> 
> 
> I don't really trust regex well enough to do this, but if it works it works! 

I did actually try that at first! The issue was that it outputted to the file everything "together". Like it wasn't every value in each line like we have it now.

PR #77, opened 2021-06-12
Print folder name alongside of the file name in table

amadejpapez:
When recursively searching there is now a folder name alongside of the file name as if there are more folders it can get confusing. Also if pyWhat searched two files with the same name, results from only one would be printed.

image


bee-san:
Any tests for this? 😄

bee-san:
good job!

PR #78, opened 2021-06-12
Changed default rarity to 0.1:1, updated docs and tests

### LINKED ISSUE #75: Move default rarity to 0.1, enable more false positive inducing regexes at 0.0
We have some regex which are very false positive trippy, we should move these to 0.0 and move the default rarity to 0.1

### LINKED ISSUE #76: Add docs for how to use tags & filters in the --help menu
Currently there is no docs :-( It might be confusing! And there's no API docs showing multiple tags either :-(

bee-san:
* Default rarity is now `0.1:1`, this allows us to include much more trigger-happy regex like key:value pairs.
* More testing!
* 2.0 release :-)
* Updated help menu so it shows more examples with multiple tags.
* Added `"URL":" to every regex so we have a more uniform experience for people using the API. (This is an API breaking change, so 3.0.0 it is!)

fixes #76 and fixes #75


pywhat/printer.py
```diff
@@ -39,7 +39,7 @@ def pretty_print(self, text: dict, text_input):
                     description = None
                     filename = key
 
-                    if "URL" in i["Regex Pattern"]:
```
bee-san, line 42: URL breaking change

pywhat/distribution.py
```diff
@@ -21,7 +21,8 @@ def __init__(self, filters_dict: Optional[dict] = None):
 
         self._dict["Tags"] = CaseInsensitiveSet(filters_dict.setdefault("Tags", tags))
         self._dict["ExcludeTags"] = CaseInsensitiveSet(filters_dict.setdefault("ExcludeTags", set()))
-        self._dict["MinRarity"] = filters_dict.setdefault("MinRarity", 0)
+        # We have regex with 0 rarity which trip false positive alarms all the time
+        self._dict["MinRarity"] = filters_dict.setdefault("MinRarity", 0.1)
```
bee-san, line 25: API level minrarity

pywhat/what.py
```diff
@@ -53,7 +53,7 @@ def parse_options(rarity, include_tags, exclude_tags):
 )
 @click.argument("text_input", required=True)
 @click.option("-t", "--tags", is_flag=True, expose_value=False, callback=print_tags, help="Show available tags and exit.")
-@click.option("-r", "--rarity", help="Filter by rarity. This is in the range of 0:1. To filter only items past 0.5, use 0.5: with the colon on the end.")
+@click.option("-r", "--rarity", help="Filter by rarity. Rarity is how unlikely something is to be a false-positive. The higher the number, the more unlikely. This is in the range of 0:1. To filter only items past 0.5, use 0.5: with the colon on the end. Default 0.1:1", default="0.1:1")
```
bee-san, line 56: My hope with `default="0.1:1"` was that it'd display in the help menu. Sadly it does not, but I think it should and it might in the future? or it might currently and i broke my local click install? :shrug:

amadejpapez (approved):

PR #79, opened 2021-06-12
only_text is set to True by default, updated tests

amadejpapez:
- `only_text` is now by default `True`, so if a user wants API to scan a file/folder, they will need to set it to `False`
- updated my recursion test - it worked really well, but it would always succeed
- added a new test that is checking if all keys are present in each JSON object - it showed that 3 were missing `plural_name`

### DIFF
--- a/pywhat/what.py
+++ b/pywhat/what.py
@@ -126,7 +126,7 @@ class What_Object:
         """
         Returns a Python dictionary of everything that has been identified
         """
-        return self.id.identify(text)
+        return self.id.identify(text, only_text=False)
 
 
 if __name__ == "__main__":
--- a/tests/test_identifier.py
+++ b/tests/test_identifier.py
@@ -1,14 +1,12 @@
-import json
 import re
 
 from pywhat import identifier
 from pywhat.distribution import Distribution
-from pywhat.helper import Keys
+from pywhat.helper import Keys, load_regexes
 
 
 def test_check_keys_in_json():
-    with open("pywhat/Data/regex.json", "r") as file:
-        database = json.load(file)
+    database = load_regexes()
 
     for entry in database:
         keys = list(entry.keys())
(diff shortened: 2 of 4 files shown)


tests/test_identifier.py
```diff
@@ -1,8 +1,28 @@
+import json
+import re
+
 from pywhat import identifier
 from pywhat.distribution import Distribution
 from pywhat.helper import Keys
 
 
+def test_check_keys_in_json():
+    with open("pywhat/Data/regex.json", "r") as file:
```
piatrashkakanstantinass, line 10: We have helper function for that, could you use it?

amadejpapez: Oh I didn't know about that! It uses the helper function now. Thank you :D

pywhat/what.py
```diff
@@ -126,7 +126,7 @@ def what_is_this(self, text: str) -> dict:
         """
         Returns a Python dictionary of everything that has been identified
         """
-        return self.id.identify(text)
+        return self.id.identify(text, only_text=False)
```
bee-san, line 129: Am I reading this correctly? There is no way to set `only_text=True` in this function? 

This means:
https://github.com/bee-san/pyWhat/blob/a4954ef6c3531c5ce87169de13b829d1fea68746/pywhat/what.py#L114-L118

For CLI input you never have only text? it will always check for a file? Might be worth making an argument for this?

tests/test_identifier.py
```diff
@@ -1,6 +1,24 @@
...
+        keys = list(entry.keys())
+        entry_name = entry["Name"]
+
+        assert "Name" in keys, entry_name
+        assert "Regex" in keys, entry_name
+        assert "plural_name" in keys, entry_name
+        assert "Description" in keys, entry_name
+        assert "Rarity" in keys, entry_name
+        assert "URL" in keys, entry_name
+        assert "Tags" in keys, entry_name
```
bee-san, line 21: noiceeeee

bee-san (approved):

PR #82, opened 2021-06-16
Add Placekey universal identifier regex

chrisbergeron:
Identify a Placekey Geolocation UID (ex.  223-227@5vg-7gq-tvz  )

Placekey URL: https://placekey.io/


bee-san:
Sure, we can add this! Heads up, we'll need to add tests to this. You can either do that by following the testing advice in the wiki, or wait until one of us checks out and adds tests to this :)

PR #85, opened 2021-06-17
add unix timestamp support

Granitosaurus:
Hello,
for my first contributions I've picked something simple: unix epoch timestamps:

```
$ poetry run pywhat 1623914861

Possible Identification
┏━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Matched Text ┃ Identified as         ┃ Description                                             ┃
┡━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ 1623914861   │ Recent Unix Timestamp │ Seconds elapsed since unix epoch: 1970, between year    │
│              │                       │ 2001 and 2286                                           │
├──────────────┼───────────────────────┼─────────────────────────────────────────────────────────┤
│ 1623914861   │ Phone Number          │ None                                                    │
├──────────────┼───────────────────────┼─────────────────────────────────────────────────────────┤
│ 1623914861   │ Unix Timestamp        │ Seconds elapsed since unix epoch: 1970                  │
└──────────────┴───────────────────────┴─────────────────────────────────────────────────────────┘
```

and milliseconds:

```
$ poetry run pywhat 1623914861000

Possible Identification
┏━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Matched Text  ┃ Identified as                     ┃ Description                                ┃
┡━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ 1623914861000 │ Recent Unix Millisecond Timestamp │ Milliseconds elapsed since unix epoch:     │
│               │                                   │ 1970, between year 2001 and 2286           │
├───────────────┼───────────────────────────────────┼────────────────────────────────────────────┤
│ 1623914861000 │ Phone Number                      │ None                                       │
├───────────────┼───────────────────────────────────┼────────────────────────────────────────────┤
│ 1623914861000 │ Unix Millisecond Timestamp        │ Milliseconds elapsed since unix epoch:     │
│               │                                   │ 1970                                       │
└───────────────┴───────────────────────────────────┴────────────────────────────────────────────┘
```

I think it would be great to support non-regex identifiers as this could have been done much more gracefully in python with `datetime` module. Nevertheless here's regex based implementation! :)


bee-san:
> I think it would be great to support non-regex identifiers 
🔜 😉

bee-san:
@amadejpapez  @piatrashkakanstantinass  what do you think of the rarity here? `0.8`? :)

Granitosaurus:
There's another caveat here that timestamps in `1970-1973` not gonna get matched since 9+ character timestamps start only from 1973-03-04~.  
I think we could reduce it to 8 which would be min `19070-04-28+`, those first 5 months though are hard to reach without introducing a lot of false-positives

bee-san:
@Granitosaurus thanks for the heads up! We'll add it to the tech-debt register (when I make one!) and when we introduce cooler checks we can do stuff like this :) I think the checks are a github issue somewhere 😓

Granitosaurus:
reduced rarity to 0.5 as recommended.

tests/test_regex_identifier.py
```diff
@@ -386,3 +389,28 @@ def test_arn4():
     r = regex_identifier.RegexIdentifier()
     res = r.check(["arn:aws:s3:::my_corporate_bucket/Development/*"])
     assert "ARN" in str(res)
+
+
+def test_unix_timestamp():
+    r = regex_identifier.RegexIdentifier()
+    ts_from_ymd = lambda ymd: int(datetime.strptime(ymd, '%Y-%m-%d').timestamp())
+
+    res = r.check([str(ts_from_ymd('2020-01-01'))])
```
SkeletalDemise, line 398: What's the point of calculating the UNIX timestamps in the tests? It would be simpler and save milliseconds of time if you hardcode them in. So instead do `res = r.check(["1623914861"])` as an example.

bee-san: I would argue that although tests are meant to be entirely repeatable, this has a higher chance of finding bugs (due to the test changing) than without?

Granitosaurus: I implemennted timestamp calcullation from strings because it's much more human readable than random strings of digits. I don't think saving 1 ms from test runs is worth the readability loss.

SkeletalDemise: I would argue that it's less readable actually. You can identify that it's a UNIX timestamp by the test name. For example the Bitcoin address tests are also not readable if you don't know what a Bitcoin address looks like but you know from the test name that it's a Bitcoin address. Also all of our other tests are strings like `1623914861`

pywhat/Data/regex.json
```diff
@@ -426,6 +427,28 @@
           "Identifiers",
           "Credentials"
        ]
+    },
+      {
+      "Name": "Recent Unix Timestamp",
+      "Regex": "^[0-9]{10}$",
+      "plural_name": false,
+      "Description": "Seconds elapsed since unix epoch: 1970, between year 2001 and 2286",
```
bee-san, line 435: @SkeletalDemise  @piatrashkakanstantinass @amadejpapez  how do we feel about these descriptions? We generally try to avoid them but I can understand Unix Timestamp is confusing.

Just wanted to get ya thoughts :)

amadejpapez: I think it should stay because as you said Unix Timestap is confusing, so this description might come in useful :)

piatrashkakanstantinass: I think we should actually write the exact time in description. e.g. phone codes

tests/test_regex_identifier.py
Granitosaurus: sure, up to the maintainers I guess! Generally having only machine readable literals in your code is just a bad practise, maybe it's not that bad for tests. Lemme know if this should be changed.

SkeletalDemise: Yeah I think you should change it. When I first looked at it, I didn't know what it was. It makes it more complex than if you just made them into strings.

bee-san:
Merge conflicts! :(

pywhat/Data/regex.json
```diff
@@ -178,6 +178,7 @@
           "IPv6"
        ]
     },
+
```
piatrashkakanstantinass, line 181: pls remove this

pywhat/Data/regex.json
```diff
@@ -679,6 +680,28 @@
           "Credentials",
           "Phone Number"
        ]
+    },
+      {
```
piatrashkakanstantinass, line 684: { should be on the same level as }

tests/test_regex_identifier.py
```diff
@@ -475,6 +478,36 @@ def test_arn4():
...
+
+    res = r.check(["1577836800000"])  # 2020-01-01
+    keys = [m['Regex Pattern']['Name'] for m in res]
+    assert "Unix Millisecond Timestamp" in keys
+    assert "Recent Unix Millisecond Timestamp" in keys
+
+    res = r.check(["94694400000"])  # 1973-01-01
+    keys = [m['Regex Pattern']['Name'] for m in res]
+    assert "Unix Millisecond Timestamp" in keys
+    assert "Recent Unix Millisecond Timestamp" not in keys
```
piatrashkakanstantinass, line 508: Can you separate groupse to be in different test functions? Like `test_unix_timestamp2()` and so on

piatrashkakanstantinass (changes requested):

tests/test_regex_identifier.py
Granitosaurus: separated 👌

pywhat/Data/regex.json
Granitosaurus: fixed

pywhat/Data/regex.json
Granitosaurus: fixed

tests/test_regex_identifier.py
```diff
@@ -475,6 +478,47 @@ def test_arn4():
     res = r.check(["arn:aws:s3:::my_corporate_bucket/Development/*"])
     assert "ARN" in str(res)
 
+
+def test_unix_timestamp1():
```
piatrashkakanstantinass, line 482: Pls remove 1 here and it will be good to go

piatrashkakanstantinass (changes requested):

tests/test_regex_identifier.py
Granitosaurus: adjusted.

piatrashkakanstantinass (approved):

piatrashkakanstantinass:
@Granitosaurus merge conflicts 😢

amadejpapez:
Should be good to go now

PR #86, opened 2021-06-17
Add support for boundaryless regex

piatrashkakanstantinass:
ToDo:

- [x] Add `Filter` class
- [x] Write tests for `Filter` class
- [x] Add suport for boundaryless regex using `Filter` class
- [x] Write some more tests
- [x] Update README
- [x] Update CLI
- [x] Add flag that disables boundaryless
- [x] Replace `re.search()`


piatrashkakanstantinass:
So far, I have added a `Filter` class. It behaves just like the `Distribution` class, although it does not store regexes.
Now, you can initialize Distribution using either another Distribution, a `Filter` or a filters_dict. The same applies to the `Filter` class.

bee-san:
This is **huge**!!!!!!!!!!!!!

pywhat/regex_identifier.py
```diff
@@ -4,21 +4,32 @@
...
+            boundaryless = Filter({"Tags": []})
         matches = []
 
         for string in text:
-            for reg in distribution.get_regexes():
-                matched_regex = re.search(reg["Regex"], string, re.UNICODE)
+            for reg in dist.get_regexes():
+                regex = reg["Regex"]
+                if reg in boundaryless:
+                    regex = regex.removeprefix("^").removesuffix("$")
```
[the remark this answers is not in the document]
piatrashkakanstantinass: Right! Didn't think about it.

bee-san:
forgot to say maybe, but this should integrate with filter system so you can filter say:

only have boundaries on regex below 1.0 rarity etc :))

piatrashkakanstantinass:
> forgot to say maybe, but this should integrate with filter system so you can filter say:
> 
> only have boundaries on regex below 1.0 rarity etc :))

This one is already done.

pywhat/Data/regex.json
```diff
@@ -37,7 +37,7 @@
     },
     {
        "Name": "TryHackMe Flag Format",
-       "Regex": "(?i)thm{.*}|tryhackme{.*}",
+       "Regex": "(?i)^(thm{.*}|tryhackme{.*})$",
```
amadejpapez, line 40: This is a really nice way to avoid in-between regex! 👀

amadejpapez: There seem to be more regexes with in-between boundaries (URL, Coordinates, Discover Card, Visa, Solo, Switch).

`Amazon Web Services Simple Storage (AWS S3) URL` has in between boundaries and also seems to match to the whole line - it should at least stop if there is a space:
`(?i)^(?:https?:\/\/s3\.amazonaws\.com\/+[^\s]+|[a-z0-9_-]*\.s3\.amazonaws\.com\/[^\s]+)$`

bee-san: Would it not perhaps be easier to add a boundaryless regex: key to the data file?

piatrashkakanstantinass: Maybe regexes should be changed to ^(pattern)$. Simple way to avoid problems with boundaries

amadejpapez: > Maybe regexes should be changed to ^(pattern)$. Simple way to avoid problems with boundaries

If we change all regexes to this format it should work yeah, and I don't think there is a need for a separate boundaryless key entry.

bee-san:
README left 😱

pywhat/filter.py
```diff
@@ -99,3 +81,54 @@ def __ior__(self, other):
...
+
+    def _filter(self):
+        self._regexes = load_regexes()
+        temp_regexes = []
+        min_rarity = self["MinRarity"]
+        max_rarity = self["MaxRarity"]
+        for regex in self._regexes:
+            if regex in self:
+                temp_regexes.append(regex)
+
```
bee-san, line 130: nice refactor 👍🏻

pywhat/regex_identifier.py
```diff
@@ -4,23 +4,34 @@
...
         matches = []
 
         for string in text:
-            for reg in distribution.get_regexes():
-                matched_regex = re.search(reg["Regex"], string, re.UNICODE)
+            for reg in dist.get_regexes():
+                regex = reg["Regex"]
+                if reg in boundaryless:
+                    regex = re.sub(r"(?<!\\)\^(?![^\[\]]*(?<!\\)\])", "", regex)
+                    regex = re.sub(r"(?<!\\)\$(?![^\[\]]*(?<!\\)\])", "", regex)
```
bee-san, line 32: Why not make a distribution with boundaryless regex? :) Would be faster over repeated calls via API :)

pywhat/what.py
```diff
@@ -68,13 +68,41 @@ def parse_options(rarity, include_tags, exclude_tags):
     help="Filter by rarity. Rarity is how unlikely something is to be a false-positive. The higher the number, the more unlikely. This is in the range of 0:1. To filter only items past 0.5, use 0.5: with the colon on the end. Default 0.1:1",
     default="0.1:1",
 )
-@click.option("-i", "--include_tags", help="Only print entries with included tags.")
-@click.option("-e", "--exclude_tags", help="Exclude tags.")
+@click.option("-i", "--include", help="Only print entries with included tags.")
+@click.option("-e", "--exclude", help="Exclude tags.")
```
bee-san, line 72:
```diff
- @click.option("-i", "--include", help="Only print entries with included tags.")
- @click.option("-e", "--exclude", help="Exclude tags.")
+ @click.option("-i", "--include", help="Only show matches with these tags")
- @click.option("-e", "--exclude", help="Disclude matches with these tags.")

File: pywhat/what.py [hunk 6]
```diff
@@ -68,13 +68,41 @@ def parse_options(rarity, include_tags, exclude_tags):
...
-@click.option("-e", "--exclude_tags", help="Exclude tags.")
+@click.option("-i", "--include", help="Only print entries with included tags.")
+@click.option("-e", "--exclude", help="Exclude tags.")
 @click.option("-o", "--only-text", is_flag=True, help="Do not scan files or folders.")
 @click.option("-k", "--key", help="Sort by the specified key.")
 @click.option("--reverse", is_flag=True, help="Sort in reverse order.")
+@click.option(
+    "-br",
+    "--boundaryless-rarity",
+    help="Same as --rarity but for boundaryless.",
```
bee-san, line 79: Why can't we just use `--rarity` for this? :)

pywhat/what.py
```diff
@@ -68,13 +68,41 @@ def parse_options(rarity, include_tags, exclude_tags):
...
+    default="0.1:1",
+)
+@click.option(
+    "-bi", "--boundaryless-include", help="Same as --include but for boundaryless."
+)
+@click.option(
+    "-be", "--boundaryless-exclude", help="Same as --exclude but for boundaryless."
+)
+@click.option(
+    "-db", "--disable-boundaryless", is_flag=True, help="Disable boundaryless mode."
```
bee-san, line 89:
```
help="Disable boundaryless mode, allows you to search within strings"
```

pywhat/what.py
```diff
@@ -68,13 +68,41 @@ def parse_options(rarity, include_tags, exclude_tags):
...
+    help="Same as --rarity but for boundaryless.",
+    default="0.1:1",
+)
+@click.option(
+    "-bi", "--boundaryless-include", help="Same as --include but for boundaryless."
+)
+@click.option(
+    "-be", "--boundaryless-exclude", help="Same as --exclude but for boundaryless."
+)
+@click.option(
```
bee-san, line 88: Is boundaryless on by default?

pywhat/what.py
```diff
@@ -68,13 +68,41 @@ def parse_options(rarity, include_tags, exclude_tags):
...
+    exclude,
+    only_text,
+    key,
+    reverse,
+    boundaryless_rarity,
+    boundaryless_include,
+    boundaryless_exclude,
+    disable_boundaryless,
+    json,
+):
```
bee-san, line 105: Would we want to switch to `**kwargs`` now? :)

pywhat/what.py
```diff
@@ -141,8 +169,14 @@ def main(text_input, rarity, include_tags, exclude_tags, only_text, key, reverse
         * what 'this/is/a/path'
 
     """
-
-    what_obj = What_Object(parse_options(rarity, include_tags, exclude_tags))
+    dist = Distribution(create_filter(rarity, include, exclude))
+    if disable_boundaryless:
+        boundaryless = Filter({"Tags": []})
```
bee-san, line 174: Why are we setting this to nothing? Can we have a comment explaining the rational please? :)

pywhat/what.py
```diff
@@ -166,11 +202,19 @@ class What_Object:
     def __init__(self, distribution):
         self.id = identifier.Identifier(dist=distribution)
 
-    def what_is_this(self, text: str, only_text: bool, key, reverse) -> dict:
+    def what_is_this(
+        self, text: str, only_text: bool, key, reverse: bool, boundaryless: Filter
```
bee-san, line 206: Nice use of type hints!

scripts/test_boundary_removal.py
```diff
@@ -0,0 +1,12 @@
...
+import re
+
+from pywhat.helper import load_regexes
+
+regexes = load_regexes()
+for regex in regexes:
+    regex["Regex"] = re.sub(r"(?<!\\)\^(?![^\[\]]*(?<!\\)\])", "", regex["Regex"])
+    regex["Regex"] = re.sub(r"(?<!\\)\$(?![^\[\]]*(?<!\\)\])", "", regex["Regex"])
+
+print(json.dumps(regexes, indent=3))
```
bee-san, line 12: You can run this as a test in CI using GitHub Actions btw if you wanted to :)

bee-san (changes requested):
Some great stuff :fire:

pywhat/what.py
piatrashkakanstantinass: Yes, in CLI with rarity 0.1:1

scripts/test_boundary_removal.py
piatrashkakanstantinass: I created it for debug purposes only, so no

pywhat/what.py
piatrashkakanstantinass: Because --rarity is used for filtration purposes and disables/enables regex fully, while --boundaryless-rarity disables/enables boundaryless mode

pywhat/what.py
piatrashkakanstantinass: Umm, '-db' does not allow to search within strings🥲

pywhat/what.py
bee-san: Can we make that clear in the help menu somehow? :)

piatrashkakanstantinass: Is it clear now?

piatrashkakanstantinass:
Breaking CLI changes btw

pywhat/what.py
piatrashkakanstantinass: I cannot find any info regarding how to do this with click(

piatrashkakanstantinass:
API docs left🚀

pywhat/helper.py
```diff
@@ -29,7 +30,15 @@ def load_regexes() -> list:
     path = "Data/regex.json"
     fullpath = os.path.join(os.path.dirname(os.path.abspath(__file__)), path)
     with open(fullpath, "r", encoding="utf-8") as myfile:
-        return json.load(myfile)
+        regexes = json.load(myfile)
+    for regex in regexes:
+        regex["Boundaryless Regex"] = re.sub(
+            r"(?<!\\)\^(?![^\[\]]*(?<!\\)\])", "", regex["Regex"]
```
amadejpapez, line 36: Now that regexes are in `^(...)$` format you could again use `removepreffix` and `removesuffix` right? If `(?i)` is causing issues it can be moved after `^`.

piatrashkakanstantinass: Yes, I can, but it means that we are forcing everyone to use this new format. Removing boundaries with regex seems to be a clever way that does not have any assumptions about the regex format.

amadejpapez: Hm true. We could make a test that would check the regex format which would make it easier to ensure that regexes in the database and from PRs are correct.

pywhat/what.py
bee-san: Can you explain what it does?

pywhat/what.py
bee-san: Just change it to **kwargs, see https://github.com/Ciphey/Ciphey/blob/master/ciphey/ciphey.py#L133

pywhat/what.py
piatrashkakanstantinass: Disables boundaryless mode so inputs like jgakljgd127.0.0.1jfl will not match

piatrashkakanstantinass: I have added an explanation of what boundaryless mode is to --help

piatrashkakanstantinass:
Finally, API docs are updated

bee-san:
Let me review and then merge :)))

pywhat/helper.py
bee-san: I would also like to see tests that make sure the regex data is sorted by rarity 👀  Homebrew has similar tests but this might be overkill haha

piatrashkakanstantinass: Actually a great idea

amadejpapez: After this gets merged, I can work on tests 😄

bee-san (commented):
Looks good other than my comments :)

README.md
```diff
@@ -62,6 +62,8 @@ What about a whole **directory**? `What` can handle that too! It will **recursiv
 
 **Exporting** You can export to json using `what --json` and results can be sent directly to a file using `what --json > file.json`.
 
+**Boundaryless mode** `What` has a special mode to match identifiable information within characters. By default, it is enabled in CLI but disabled in API. Use `what --help` or refer to [API Documentation](https://github.com/bee-san/pyWhat/wiki/API) for more information.
```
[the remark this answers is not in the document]
piatrashkakanstantinass: I don't think that the **Anything** section is large enough to move it to the wiki. On the other hand, it may be simplified to just a list of all the features with zero explanation and a note about what users need to do to get CLI or API info. Your thoughts, @bee-san?

bee-san (approved):

PR #89, opened 2021-06-18
feat: add Media Access Control (MAC) Address regex

yihong0618:


Granitosaurus:
imo rarity is pretty low here - I think mac addresses unique enough looking for something higher like 0.8 or even 1.

PR #91, opened 2021-06-19
Adding AWS credential types

belialboy:
Adding AWS Access Key and AWS Secret Access Key to the supported regex patterns


bee-san:
Thank you for this! This needs tests, feel free to add them or we can do it later :)

belialboy:
My bad. I've added the tests for you.

pywhat/Data/regex.json
```diff
@@ -151,6 +151,30 @@
...
     },
+    {
+      "Name": "Amazon Web Services Access Key",
+      "Regex": "(?<![A-Z0-9])[A-Z0-9]{20}(?![A-Z0-9])",
+      "plural_name": false,
+      "Description": "Access key used to authenticate with the AWS API",
+      "Rarity": 1,
+      "URL": null,
+      "Tags": [
+         "Credentials",
```
bee-san, line 162: I just remembered!! We have an `AWS` tag, mind adding that? And if you spot any other AWS regex without it please add it :) AWS has so many (albeit, confusing) IDs that it needs its own tag haha!

bee-san: other than this, looks good to go 👍🏻

belialboy: Can do. Would you like me to add some of the other AWS types like security group, EC2 instance, organization ID? They'll be less common, but do have a fixed format.

bee-san: Sure, if you don't mind! :)

belialboy: Done... I'll see if I can think of others on my travels and will update as I go. Ok?

piatrashkakanstantinass:
Merge conflicts😢

bee-san:
If possible, please fix the merge conflicts && remove the descriptions -- we try to only include them if it's needed (I.E. the name is not descriptive of what it is)

amadejpapez:
Seems like some newly added API key formats are first getting matched as AWS ones. You can change tests to this format:
`assert "Amazon Web Services Access Key" in res` as currently test is checking only the first match.

amadejpapez:
I have fixed tests and removed `AWS Access Key ID` as I think it is the same as newly added `Amazon Web Services Access Key`.

PR #96, opened 2021-06-21
Add short-names to tags #69

### LINKED ISSUE #69: Add short-names to tags
For IP addresses we can do:
* ipv4
* ipv6

etc

This is so users can hand-select what regex they want via the tag system :D

RackReaver:
Added multiple new tags to allow for hand-selection.

Fixes #69

### DIFF
--- a/pywhat/Data/regex.json
+++ b/pywhat/Data/regex.json
@@ -49,9 +49,8 @@
        "Rarity": 1,
        "URL": "https://etherscan.io/address/",
        "Tags": [
-          "Cryptocurrency",
           "Finance",
-          "Coin Wallet",
+          "Cryptocurrency Wallet",
           "Ethereum Wallet"
        ]
     },
@@ -77,7 +76,7 @@
        "Tags": [
           "Cryptocurrency",
           "Finance",
-          "Coin Wallet",
+          "Cryptocurrency Wallet",
           "Bitcoin Wallet"
        ]
     },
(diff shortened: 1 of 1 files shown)


pywhat/Data/regex.json
```diff
@@ -112,7 +122,8 @@
        "Rarity": 1,
        "URL": null,
        "Tags": [
-          "Media"
+          "Media",
+          "YouTube Video"
```
bee-san, line 126: We should add "youtube" here

pywhat/Data/regex.json
```diff
@@ -68,7 +76,9 @@
        "URL": "https://www.blockchain.com/btc/address/",
        "Tags": [
           "Cryptocurrency",
-          "Finance"
+          "Finance",
+          "Coin Wallet",
```
bee-san, line 80: Maybe change to cryptocurrency coin wallet?

bee-san (changes requested):
Thanks so much! Looks great :)

pywhat/Data/regex.json
RackReaver: I'm good with that, wasn't sure if there was a need for "cryptocurrency". I will make the change and fix in a bit.

RackReaver: I removed `Cryptocurrency` and renamed `Coin Wallet` to `Cryptocurrency Wallet`

pywhat/Data/regex.json
RackReaver: Done.

bee-san:
thanks!!!!

PR #95, opened 2021-06-21
Update American Social Security Number.

RackReaver:
- No numbers with all zeros in any digit group
- No numbers with 666 or 900–999 (Individual Taxpayer Identification Number) in the first digit group.
- Account for potential digit group spacers `= | \+ | _ | \# | : | ; | \. | - | \s` 

References:
https://en.wikipedia.org/wiki/Social_Security_number
https://www.usrecordsearch.com/ssn.htm

PR #102, opened 2021-06-26
Update American Social Security Number.

RackReaver:
- No numbers with all zeros in any digit group
- No numbers with 666 or 900–999 (Individual Taxpayer Identification Number) in the first digit group.
- Account for potential digit group spacers `= | \+ | _ | \# | : | ; | \. | - | \s` 

#95 Re-opening failed pull request

References:
https://en.wikipedia.org/wiki/Social_Security_number
https://www.usrecordsearch.com/ssn.htm


RackReaver:
@bee-san I closed out #95 by mistake myself so I just opened a new pull request.

bee-san:
There's a merge conflict :(

RackReaver:
dam 😆 let me take a look.

RackReaver:
> There's a merge conflict :(

Should be good now

tests/test_identifier.py
```diff
@@ -44,7 +44,7 @@ def test_identifier_works3():
     out = r.identify("fixtures/file", only_text=False)
     assert (
         "Dogecoin (DOGE) Wallet Address"
-        in out["Regexes"]["file"][1]["Regex Pattern"]["Name"]
+        in out["Regexes"]["file"][2]["Regex Pattern"]["Name"]
```
bee-san, line 47: This is scary 👻

bee-san: @piatrashkakanstantinass  thoughts?

RackReaver: For context I needed to change this to get the test to pass, I didn't really look into why this test broke though.

amadejpapez: > This is scary 👻 

Thankfully we have tests, so issues like this get noticed quickly :D

@RackReaver The problem is that pyWhat finds two SSNs inside of the DOGE address. I don't think the regex can get restricted any more, so please add boundaries to the regex (so that it won't match inside of text) and revert this change in tests. :)

pywhat/Data/regex.json
```diff
@@ -703,19 +703,18 @@
...
-       "Regex": "^[0-9]{3}-[0-9]{2}-[0-9]{4}$",
-       "plural_name": false,
-       "Description": "An [#CAE4F1][link=https://en.wikipedia.org/wiki/Social_Security_number]American Identification Number[/link][/#CAE4F1]",
-       "Rarity": 0.2,
-       "URL": null,
-       "Tags": [
-          "Credentials",
-          "Password",
-          "Username",
-          "SSN"
```
amadejpapez, line 716: Please add `SSN` tag back. It came from another PR, so that's why it probably slipped through :)

RackReaver: Done.

tests/test_identifier.py
RackReaver: Done.

amadejpapez: Thank you! Looks good to me now! 😄

PR #104, opened 2021-06-28
Amazon Standard Identification Number (ASIN)

mimiflynn:


pywhat/Data/regex.json
```diff
@@ -394,6 +394,18 @@
           "Networking"
        ]
     },
+    {
+      "Name": "Amazon Standard Identification Number (ASIN)",
+      "Regex": "(?:[/dp/]|$)([A-Z0-9]{10})",
+      "plural_name": false,
+      "Description": null,
+      "Rarity": 0.5,
+      "URL": null,
```
amadejpapez, line 403: Hi! Thank you very much for contributing!

I have looked into this a little bit and seems there is a way to access the product page with only the ASIN number, so could you please add `https://amzn.com/` here? PyWhat will add the ASIN number at the end and users will be able to see what product it is for. 😄

amadejpapez (changes requested):

pywhat/Data/regex.json
```diff
@@ -394,6 +394,18 @@
           "Networking"
        ]
     },
+    {
+      "Name": "Amazon Standard Identification Number (ASIN)",
+      "Regex": "(?:[/dp/]|$)([A-Z0-9]{10})",
```
piatrashkakanstantinass, line 399: We want all regexes to have hardcoded boundaries (^$.) Can you add these? Btw, I see that some entries do not have those either. It would be fantastic if you could add boundaries.

piatrashkakanstantinass (changes requested):
Also, please format tests using black

pywhat/Data/regex.json
mimiflynn: Great idea!

pywhat/Data/regex.json
mimiflynn: I'm not super confident with regex just yet. Is there a line that can serve as an ideal example of this?

amadejpapez: I can explain it to you. For example, you want to match to the word `amazon`. By default regex should only match if `amazon` is in its own line. It should match to this line:
`amazon`
And not to `amazon` in this line:
`amazon test`

To accomplish this you need to add `^` at the beginning of regex which means start of line, and `$` at the end of regex which means the end of line.

So change your regex to this: `^(your regex)$`. Some others like SSH and API ones that were added recently also need this. It would be amazing if you could fix that. 😁

You can experiment with regex on a website like `regex101.com`. If you have any further questions, I am happy to help. 😄

mimiflynn: Thanks for the detailed explanation! I'll be sure to reach out with any questions.

mimiflynn:
I'm not quite sure whats happening: with the hardcoded boundaries `^((?:[/dp/]|$)([A-Z0-9]{10}))$`, the tests fail and without them `(?:[/dp/]|$)([A-Z0-9]{10})` they pass. I'm looking into it more, but if you have any quick insights, I'd appreciate the feedback.

Thanks!

amadejpapez:
This is happening because with hardcoded boundaries the whole match needs to be in one line with nothing else around it.

`http://www.amazon.com/Kindle-Wireless-Reading-Display-Generation/dp/B0015T963C`

Your regex is matching to only `/dp/B0015T963C`, so it doesn't match as there is something before it. If you add only this part as a test it will pass. 😄

mimiflynn:
> This is happening because with hardcoded boundaries the whole match needs to be in one line with nothing else around it.
> 
> `http://www.amazon.com/Kindle-Wireless-Reading-Display-Generation/dp/B0015T963C`
> 
> Your regex is matching to only `/dp/B0015T963C`, so it doesn't match as there is something before it. If you add only this part as a test it will pass. smile

Oh, I see! I had lost sight of main use of `pyWhat` while using regex101 and was extracting ASINs from strings instead of checking if a string is an ASIN. Updating now with new regex removing URL specific aspects.

piatrashkakanstantinass (approved):

bee-san:
> > This is happening because with hardcoded boundaries the whole match needs to be in one line with nothing else around it.
> > `http://www.amazon.com/Kindle-Wireless-Reading-Display-Generation/dp/B0015T963C`
> > Your regex is matching to only `/dp/B0015T963C`, so it doesn't match as there is something before it. If you add only this part as a test it will pass. smile
> 
> Oh, I see! I had lost sight of main use of `pyWhat` while using regex101 and was extracting ASINs from strings instead of checking if a string is an ASIN. Updating now with new regex removing URL specific aspects.

I think in bounardyless mode we do extract that string, but we need it to have boundaries in our database so we can take them away in our boundaryless mode :)

bee-san (approved):

PR #111, opened 2021-07-08
Update dev-dependencies

piatrashkakanstantinass:


pyproject.toml
```diff
@@ -8,12 +8,16 @@ documentation = "https://github.com/bee-san/pyWhat/wiki"
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = "^3.6"
+python = "^3.6.2"
```
bee-san, line 11: What happens between .6 and .6.2 that we need this? :)

piatrashkakanstantinass: black requires python 3.6.2 or higher @bee-san

SkeletalDemise: Then add it under the dev dependencies so that people can still use Python `3.6.0` if they want.

bee-san: Yeah I agree with Skeletal here, REMnux requires 3.6 :shrug: :(

bee-san (approved):

bee-san:
merge conflicts :(

SkeletalDemise (approved):

PR #114, opened 2021-07-12
Add more 'Media' related regexes

apteryxxyz:
This adds media related regexes similar to the existing YouTube Video regex, in which they match the URL of something like a social media post.

Includes:
- Twitter Status
- Twitter User
- Instagram Post
- Instagram User
- Reddit Post
- Reddit User
- Reddit Subreddit
- Discord Invite

These will match URLs, so for example 'https://twitter.com/jack/status/20/', 'https://reddit.com/r/funny', 'https://discord.com/invite/hTKzmak', etc.

### DIFF
--- a/pywhat/Data/regex.json
+++ b/pywhat/Data/regex.json
@@ -151,7 +151,7 @@
    },
    {
       "Name": "Instagram Post",
-      "Regex": "^(https?:\\/\\/(?:(www\\.)?instagram\\.com\\/p\\/)([^/?#&]{1,15})(?:\\/)?)$",
+      "Regex": "^(https?:\\/\\/(?:(www\\.)?instagram\\.com(\\/[A-Za-z0-9_.]{1,30})?\\/p\\/)([a-zA-Z0-9_-]+)(?:\\/)?)$",
       "plural_name": false,
       "Description": null,
       "Rarity": 1,
@@ -163,7 +163,7 @@
    },
    {
       "Name": "Instagram User",
-      "Regex": "^(https?:\\/\\/(?:(www\\.)?instagram\\.com\\/)([^/?#&]{1,15})(?:\\/)?)$",
+      "Regex": "^(https?:\\/\\/(?:(www\\.)?instagram\\.com\\/)([A-Za-z0-9_.]{1,30})(?:\\/)?)$",
       "plural_name": false,
       "Description": null,
       "Rarity": 1,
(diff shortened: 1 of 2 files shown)


tests/test_regex_identifier.py
```diff
@@ -448,6 +448,60 @@ def test_youtube_channel_id():
     _assert_match_first_item("YouTube Channel ID", res)
 
 
+def test_twitter():
+    r = regex_identifier.RegexIdentifier()
+    res = r.check(["https://twitter.com/jack/status/20"], boundaryless=Filter({"Tags": ["Media"]}))
```
piatrashkakanstantinass, line 453: Could you remove the boundaryless part in every test since it is not necessary?

piatrashkakanstantinass (changes requested):

tests/test_regex_identifier.py
apteryxxyz: Sure thing, done

pywhat/Data/regex.json
```diff
@@ -125,6 +125,102 @@
          "YouTube Video"
       ]
    },
+   {
+      "Name": "Twitter Status",
+      "Regex": "^(https?:\\/\\/(?:twitter\\.com\\/)(?:#!\\/)?(\\w{1,15})\\/status(?:es)?\\/(\\d+))$",
```
amadejpapez, line 130: It may be a good idea to make all of these regexes case insensitive (add `(?i)` after `^`), so they match to uppercase URLs too: HTTPS://TWITTER.COM/BEE_SEC_SAN/STATUS/1413943153732816902?S=21

apteryxxyz: Okay, although the existing YouTube Video [URL] regex doesn't have `(?i)`

amadejpapez: Make one URL uppercase and test it. If it works, you can make it case insensitive. Also please test other URLs too as these contain IDs and just to be sure that they also work when fully uppercase :D

apteryxxyz: Added `(?:)` to regexes and more tests.

bee-san:
Hey! 👋🏻  I'm not sure on how okay  I am at providing regex for specific sites when we already have a URL regex, is there any reason you'd want this? 😄  ❤️

apteryxxyz:
> Hey! 👋🏻 I'm not sure on how okay I am at providing regex for specific sites when we already have a URL regex, is there any reason you'd want this? 😄 ❤️

Mainly just to add more 'media' regexes. The existing 3 YouTube regexes are the only [social] media ones, so similar regexes for other social networks seemed to fit.

PR #119, opened 2021-07-19
Sort TLDs by length

### LINKED ISSUE #117: URL regex does not fully match every URL
For example:
image

I would like it to return both google.co and google.com. Sadly, it may be impossible considering the way regular expressions work. Thus, it would be amazing to match the longest string (`pywhat google.com/help` should return `google.com/help`). It is crucial for implementing URL subcategories properly(#51). Btw, URL regex is too long, I do not think that valid TLDs should be checked, so it may be shortened.

piatrashkakanstantinass:
Closes #117


bee-san (approved):
Nice one!!! :clap: 🥳

PR #123, opened 2021-07-27
Added and removed some of the tags

amadejpapez:
I went through the list of available tags and here are some of the changes I made:
- removed ```Pretty Good Privacy``` as it is the same as ```PGP```
- removed ```Cryptocurrency``` as it was only used for Bitcoin and there is already ```Cryptocurrency wallet``` tag
- removed some of the more broad tags as they were used only on some regexes and I think there are better more specific tags
- renamed ```Time``` to ```UNIX Timestamp```
- added ```CTF flag``` tag, instead of having them under ```Cyber Security```
- added ```IP``` tag for both IPv4 and IPv6
- added ```Finance``` tag to Laser Card Number as it was missing

### DIFF
--- a/pywhat/Data/regex.json
+++ b/pywhat/Data/regex.json
@@ -7,8 +7,7 @@
       "Rarity": 1,
       "URL": null,
       "Tags": [
-         "PGP",
-         "PGP public Key"
+         "PGP"
       ]
    },
    {
@@ -19,8 +18,7 @@
       "Rarity": 1,
       "URL": null,
       "Tags": [
-         "PGP",
-         "PGP private Key"
+         "PGP"
       ]
    },
    {
@@ -32,7 +30,7 @@
       "URL": null,
       "Tags": [
          "Credentials",
-         "SSH RSA Public Key"
+         "SSH Public Key"
       ]
    },
    {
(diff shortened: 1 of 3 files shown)


piatrashkakanstantinass:
Don't yoy think that tags like PGP Public key and PGP Private key do not make a lot of sense. These tags are just the names of entries. I would say that it is better to change these to PGP. This may be the same for some other tags aswell.

amadejpapez:
Yeah agree. Do you see any other tags like this that could be removed?

piatrashkakanstantinass:
IPv6, IPv4, MaestroCC, Mailgun. Also rename `CTF flag` to `CTF Flag`

piatrashkakanstantinass (approved):

pywhat/Data/regex.json
```diff
@@ -238,7 +225,7 @@
       "Tags": [
          "Identifiers",
          "Networking",
-         "IPv6"
```
bee-san, line 241: The reason we used specific tags here was so people could pick and choose what they wanted specifically, if we can filter on names this isn't needed :)

amadejpapez: IPv4 and IPv6 are the ones that may actually come useful, so they can be added back yeah. Are there any others?

piatrashkakanstantinass: In this case, it may be better to add individual tags to all entries.

PR #124, opened 2021-07-28
Completely remove Name-That-Hash support

amadejpapez:


pywhat/identifier.py
```diff
@@ -3,9 +3,8 @@
 from typing import Callable, Optional
 
 from pywhat.filter import Distribution, Filter
-from pywhat.helper import AvailableTags, Keys
```
bee-san, line 6: why?

bee-san (changes requested):

pywhat/identifier.py
amadejpapez: It isn't used in the file and that's why I removed it.

bee-san: oki

bee-san (approved):

SkeletalDemise (approved):

PR #127, opened 2021-08-02
Implement subcategories (WIP)

### LINKED ISSUE #51: Create "regex within regex" for subcategories
Some regex have sub-categories, for example:
* Mastercard numbers -- the first 4 digits represent what company that card belongs to.
* Phone numbers. If they start with +44 or 07, it's the UK!

As such, I propose we change our code into:

```json
    {
       "Name": "MasterCard Number",
       "Regex": "^(?:5[1-5][0-9]{2}|222[1-9]|22[3-9][0-9]|2[3-6][0-9]{2}|27[01][0-9]|2720)[0-9]{12}$",
       "plural_name": false,
       "Description": null,
       "Rarity": 0.5,
       "Tags": [
          "Credit Card",
          "Finance"
       ],
        "Children": [
            {
                "Name": "BarclayCard",
                "Regex:" "5117.*"
                Same value as the parent above
            }
        ]
    },
```

We want to keep the same object for each child in case we have anything special to add like a website or a cool description etc. It doesn't cost us much to do this, but it is helpful!

Also, our regex for the children can be much broader. Read it as:
1. When I pass the mastercard regex
2. When I pass the barclaycard regex (which just checks to see if the first few numbers are equal to 5117)
3. Return
Because the children only run when the parents do, the child regex can be lighter.

As an aside, Mastercard have very helpfully provided a CSV table of all the companies and what their Mastercard subrange is:
https://www.mastercard.us/en-us/business/issuers/get-support/simplified-bin-account-range-table.html

piatrashkakanstantinass:
Closes #51


piatrashkakanstantinass:
ToDo:

- [ ] Add optional hashmap support (only checks beginning of a match) for better speed
- [ ] Check if I can use faster JSON lib

Testing is slow as hell

pywhat/Data/regex.json
```diff
@@ -287,7 +287,13 @@
       "Tags": [
          "Identifiers",
          "Networking"
-      ]
+      ],
+      "Children": {
+	      "path": "mac_vendors.json",
+	      "entry": "Vendor(s): ",
+	      "method": "hashmap",
```
bee-san, line 294: > 	      "method": "hashmap",

This is cool

pywhat/helper.py
```diff
@@ -1,9 +1,14 @@
...
-import json
 import os.path
 import re
 from enum import Enum, auto
+from functools import lru_cache
+
+try:
+    import orjson as json
+except ImportError:
+    import json
```
bee-san, line 11: This is cool actually

pywhat/helper.py
```diff
@@ -26,18 +31,31 @@ class InvalidTag(Exception):
...
-    path = "Data/regex.json"
-    fullpath = os.path.join(os.path.dirname(os.path.abspath(__file__)), path)
+def read_json(path: str):
+    fullpath = os.path.join(os.path.dirname(os.path.abspath(__file__)), "Data/" + path)
     with open(fullpath, "rb") as myfile:
-        regexes = json.load(myfile)
+        return json.loads(myfile.read())
+
+
+@lru_cache()
```
bee-san, line 40: Why do we cache this? Do we call this more than once? :)

pywhat/identifier.py
```diff
@@ -15,7 +15,7 @@ def __init__(
         dist: Optional[Distribution] = None,
         key: Callable = Keys.NONE,
         reverse=False,
-        boundaryless: Optional[Filter] = None
+        boundaryless: Optional[Filter] = None,
```
bee-san, line 18: Nice catch

bee-san (approved):
🔥

pywhat/helper.py
piatrashkakanstantinass: Yes, in API mode if many distributions/identifiers are being used. +it *greately* improves the speed of tests.

bee-san:
Feel free to merge whenever u want

PR #128, opened 2021-08-06
Make searching filenames for matches optional

amadejpapez:
Previously PyWhat always added the name of the file to the search. This PR makes it optional, so if someone wants PyWhat to also search filenames for any matches, they can now do it with `-if` command.

### DIFF
--- a/pywhat/what.py
+++ b/pywhat/what.py
@@ -90,11 +90,10 @@ def create_filter(rarity, include, exclude):
 )
 @click.option("--json", is_flag=True, help="Return results in json format.")
 @click.option(
-    "-fn",
-    "--search-filenames",
+    "-if",
+    "--include-filenames",
     is_flag=True,
-    help="Alongside of file contents also search filenames for any possible matches.",
-    default=False,
+    help="Search filenames for possible matches."
 )
 def main(**kwargs):
     """
@@ -211,7 +210,7 @@ def main(**kwargs):
         key,
         kwargs["reverse"],
         boundaryless,
-        kwargs["search_filenames"],
+        kwargs["include_filenames"],
     )
 
     p = printer.Printing()
@@ -234,7 +233,7 @@ class What_Object:
         key,
         reverse: bool,
         boundaryless: Filter,
-        search_filenames: bool,
+        include_filenames: bool,
     ) -> dict:
         """
         Returns a Python dictionary of everything that has been identified
@@ -245,7 +244,7 @@ class What_Object:
             key=key,
             reverse=reverse,
             boundaryless=boundaryless,
-            search_filenames=search_filenames
+            include_filenames=include_filenames
         )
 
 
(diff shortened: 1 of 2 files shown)


pywhat/what.py
```diff
@@ -89,6 +89,13 @@ def create_filter(rarity, include, exclude):
     "-db", "--disable-boundaryless", is_flag=True, help="Disable boundaryless mode."
 )
 @click.option("--json", is_flag=True, help="Return results in json format.")
+@click.option(
+    "-fn",
+    "--search-filenames",
+    is_flag=True,
+    help="Alongside of file contents also search filenames for any possible matches.",
+    default=False,
```
piatrashkakanstantinass, line 97: Why do you use `default` here?

piatrashkakanstantinass (changes requested):
API docs should be updated after this gets merged.

pywhat/what.py
```diff
@@ -89,6 +89,13 @@ def create_filter(rarity, include, exclude):
     "-db", "--disable-boundaryless", is_flag=True, help="Disable boundaryless mode."
 )
 @click.option("--json", is_flag=True, help="Return results in json format.")
+@click.option(
+    "-fn",
+    "--search-filenames",
```
bee-san, line 94: This name is confusing, maybe `--include-filenames`?

pywhat/what.py
```diff
@@ -89,6 +89,13 @@ def create_filter(rarity, include, exclude):
     "-db", "--disable-boundaryless", is_flag=True, help="Disable boundaryless mode."
 )
 @click.option("--json", is_flag=True, help="Return results in json format.")
+@click.option(
+    "-fn",
+    "--search-filenames",
+    is_flag=True,
+    help="Alongside of file contents also search filenames for any possible matches.",
```
bee-san, line 96:
```diff
- Alongside of file contents also search filenames for any possible matches.
+ Search filenames for possible matches.
```

bee-san (changes requested):

pywhat/what.py
amadejpapez: True I will rename it. Is short `fn` okay?

bee-san: It might be confusing since the short is different from the long, if you only knew the long I think you'd expect it to be `-if` :)

piatrashkakanstantinass (approved):

pywhat/what.py
```diff
@@ -69,7 +69,7 @@ def create_filter(rarity, include, exclude):
     default="0.1:1",
 )
 @click.option("-i", "--include", help="Only show matches with these tags.")
-@click.option("-e", "--exclude", help="Disclude matches with these tags.")
+@click.option("-e", "--exclude", help="Exclude matches with these tags.")
```
bee-san, line 72: NICEE!!!!!

bee-san (approved):

bee-san:
@amadejpapez you can merge if you want!! normally in industry ✨  when ew approve something, its up to you to merge it as you may want to add some things last minute etc :)

PR #129, opened 2021-08-07
Better regex for BTC, BCH and XMR wallets

amadejpapez:
- stricter BTC regex and support for newer BECH32 format (new tests for all three formats)
- stricter BCH and XMR regexes (Thank you Nick Carr!)


bee-san (approved):

bee-san:
love u

Issue #152, opened 2021-10-02
Automatically fill in the Curl webhook


bee-san:
1. We have exploits which use CURL like this https://github.com/bee-san/pyWhat/blob/main/pywhat/Data/regex.json#L137
2. `curl -s -X POST -H \"Content-type: application/json\" -d '{\"text\":\"\"}' \"SLACK_WEBOOK_URL_HERE\"`
3. Where it says `SLACK_WEBOOK_URL_HERE` automatically fill in the webhook (provided because the regex found it)

PR #157, opened 2021-10-02
Automatically fill in the Curl webhook

colinodell:
Implements #152

Not the prettiest solution but it seems to work fine, so long as the exploit uses a placeholder that's in all-caps and ends with `_HERE`.

### DIFF
--- a/pywhat/Data/regex.json
+++ b/pywhat/Data/regex.json
@@ -650,7 +650,7 @@
       "Regex": "(?i)^(NRRA-[a-f0-9]{42})$",
       "plural_name": false,
       "Description": null,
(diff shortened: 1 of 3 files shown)


piatrashkakanstantinass:
This feature should be implemented on the API level, so all calls to Identifier.identify() return formatted exploit. Check `regex_identifier.py`.

piatrashkakanstantinass:
Failing tests😢

colinodell:
Yep, that's what I get for not running the test suite locally before pushing :facepalm: should be fixed now!

I also noticed that the Slack exploits included a redundant `xoxp-` prefix immediately before the `TOKEN_HERE`, so I removed that and added a missing test.

piatrashkakanstantinass (approved):

piatrashkakanstantinass:
That is amazing! Thank you a lot @colinodell 😀

bee-san:
Thanks for thiis! <3

PR #160, opened 2021-10-02
Added Windows' File Name Regex match and YouTube Channel ID

Latonis:
Added both the Regex to the json and testing.

### DIFF
--- a/pywhat/Data/regex.json
+++ b/pywhat/Data/regex.json
@@ -1536,5 +1536,17 @@
       "Tags": [
          "Credentials"
       ]
-   }
+   },
+	{
+     "Name": "Windows File Name",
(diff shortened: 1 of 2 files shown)


bee-san:
Hey 👋🏻  Thanks for making a PR! :) 

We already have Youtube channel ID here: https://github.com/bee-san/pyWhat/blob/674bb9a74907ff530f78b185704d3746c5059637/pywhat/Data/regex.json#L1029

image

And I don't think your regex works for Windows files? :)

Latonis:
Hi! I think i might have included the regex incorrectly, I can take a look this week to resolve. Sorry for the late response :)

Latonis:
image
I just tested again, I had too many escapes. 
image

This is the correct regex pattern:
`^(?!(?:CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])(?:\.[^.]*)?$)[^<>:\"/\\|?*\x00-\x1F]*[^<>:\"/\\|?*\x00-\x1F\.]$`

Issue #130, opened 2021-08-16
Add Turkey Regex Patterns


bee-san:
This repo:
https://translate.google.com/translate?sl=auto&tl=en&u=https://github.com/mgulener/turkiye-regex-kaliplari

Is described as:

> Regex codes of some patterns used in Turkey. We can add some of these which **we don't have already**.

You can follow this guide:
https://github.com/bee-san/pyWhat/wiki/Adding-your-own-Regex

to add your own Regex! :) <3

piatrashkakanstantinass:
Btw, @bee, we need different format support (aka multiple regexes). That is useful for phone numbers and a lot of other stuff.

sonniki:
I'll take this one if you don't mind :)

PR #169, opened 2021-10-04
Added Turkish regexes

sonniki:
Solves #130 

The Turkish regex repo also had the regexes for the phone numbers and credit cards, but they are already included here, even with bigger coverage and specificity.


bee-san:
Great tests!!! 👏🏻  Thanks so much for this :)

PR #172, opened 2021-10-05
Fix nox tests

SkeletalDemise:
Fixes nox file to not be restricted to only using Python 3.8 and Python 3.7. This fixes the Python 3.6 and Python 3.9 CI checks.


bee-san:
all ya tests fail also please squash commits :)

PR #173, opened 2021-10-05
Regex Tests: Added Hawk API test and Fixed Discord WebHook Regex

0xmostafam:
Hello 👋
- Added regex tests for Hawk API and Discord Webhook
- Correct me if i am wrong but when i searched for Discord Webhooks, it was actually discord.com/api/webhooks/ not discordapp.com/api/webhooks/

### DIFF
--- a/pywhat/Data/regex.json
+++ b/pywhat/Data/regex.json
@@ -731,7 +731,7 @@
    },
    {
       "Name": "Discord Webhook",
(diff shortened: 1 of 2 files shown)


bee-san:
Woo!! Thanks so much :D 

> Correct me if i am wrong but when i searched for Discord Webhooks, it was actually discord.com/api/webhooks/ not discordapp.com/api/webhooks/

Probably both, I think they just redirect now :)

bee-san (approved):

PR #175, opened 2021-10-05
Added / fixed tests and regex 

0xmostafam:
Hello Again 👋

- Added tests and regex for Visual Studio Token and NPM Token
- Correct me if i am wrong but Bitly regex was incorrect because according to this video, https://www.youtube.com/watch?v=2rL4K4YOop8 and i even registered for a account to try it, the secret key was combosed of 40 hex characters
- I couldn't use `_assert_match_first_item` because Visual Studio Token and Bitly now has the same regex, and i noticed that there was an unused function called `_assert_match_in_items` so i rewrote it and used it for both of them

and if you want to me to refactor the tests so that all of them uses either `_assert_match_in_items` and `_assert_match_first_item` only i can do it


pywhat/Data/regex.json
```diff
@@ -333,6 +333,19 @@
          "Bug Bounty"
       ]
    },
+   {
+      "Name": "NPM Token",
```
bee-san, line 337: In your personal opinion, should we rename it to:

```suggestion
      "Name": "Node Package Manager (NPM) Token",
```

? Or is NPM so popular that people know it by acronym?

bee-san (commented):
Some great work here 👏👏👏👏 Thank you!!! 💖💖💖

tests/test_regex_identifier.py
```diff
@@ -24,7 +24,10 @@ def test_regex_successfully_parses():
 
 def _assert_match_in_items(name, res):
```
piatrashkakanstantinass, line 25: Can you change this to:
```py
def _assert_match_in_items(name, res):
    assert any(name in i["Regex Pattern"]["Name"] for i in res)
```
?

0xmostafam: sure

bee-san (approved):

PR #177, opened 2021-10-06
Added support for Notion Notes and API keys

alb:
This PR adds support for [Notion's](https://www.notion.so/):
- Individual and Team Notes
- API Keys for Notion Integrations

### DIFF
--- a/pywhat/Data/regex.json
+++ b/pywhat/Data/regex.json
@@ -867,9 +867,21 @@
    },
    {
       "Name": "Notion Note URI",
-      "Regex": "(?i)^(https:\/\/www.notion.so\/[A-Za-z0-9-_]+\/.*-.{32})$",
+      "Regex": "(?i)^(https:\/\/www\\.notion\\.so\/[A-Za-z0-9-_]+\/.*-.{32})$",
       "plural_name": false,
-      "Description": "A note written on notion.so",
+      "Description": "A note written on notion.so.",
+      "Rarity": 1,
+      "URL": "https://www.notion.so/",
+      "Tags": [
+         "Bug Bounty",
+         "URL"
+      ]
+   },
+   {
+      "Name": "Notion Team Note URI",
+      "Regex": "(?i)^(https:\/\/[A-Za-z0-9-_]+\\.notion\\.site\/.*-.{32})$",
+      "plural_name": false,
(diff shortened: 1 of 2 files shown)


alb:
@bee-san quick question:
Should I squash the commits into one or is it fine as is?

pywhat/Data/regex.json
```diff
@@ -865,6 +865,30 @@
          "AWS"
       ]
    },
+   {
+      "Name": "Notion Note URI",
+      "Regex": "(?i)^(https:\/\/www\\.notion\\.so\/[A-Za-z0-9-_]+\/.*-.{32})$",
+      "plural_name": false,
+      "Description": "A note written on notion.so.",
+      "Rarity": 1,
+      "URL": "https://www.notion.so/",
```
bee-san, line 874: Hey! Some notes:
* We try to avoid descriptions unless they are absolutely necessary. Like social security number. To an American that means something, but to others it does not :(  I think the description is just reiterating the name :D 
* URLs do not work like this, they append the matched regex to the end.

For example, say your Notion URL is  https://notion.so/kdkaddjad

With this URL key, it'll be printed as `https://www.notion.so/https://notion.so/kdkaddjad` 

We use this to send users to a given website if it helps them. For example. if it's a Bitcoin address we can send them to:
```
Matched text: bc1q52rrnrvjr75elkj5q5nv006h8hyc5jfknndzdn
URL: https://www.blockchain.com/btc/address/

Prints: https://www.blockchain.com/btc/address/bc1q52rrnrvjr75elkj5q5nv006h8hyc5jfknndzdn
```

bee-san (changes requested):

bee-san:
You're fine with not squashing :)

pywhat/Data/regex.json
alb: My bad, I should have given a wiki a longer look!
Description is gone now and the regex should capture the string correctly.
The `Team Note URI` regex doesn't have a link so I removed the URL and I left the regex as is since the `subdomain.notion.site` combination will always change depending on the URL.

bee-san:
Can you delete the URLs on the other ones too please? :)

alb:
@bee-san hope I got this right this time around :sweat_smile: 
Now the Notion URI only matches the identifier so it *should* be appended correctly.

PR #190, opened 2021-10-09
Add TOTP URIs

PabloLec:
This commit adds TOTP URIs regex and related tests.
You might find TOTP URIs in Google Authenticator, Authy, Bitwarden, etc. backups. Or just anywhere if you're doing Android/iOS pentesting. And with its base-32 key, TOTP is all yours !

I might add some other pentesting related regex in the next few days :)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 90%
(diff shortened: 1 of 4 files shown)


PabloLec:
Ok, just pushed the above edits and added a more complex test, especially to test arguments position swapping inside the URI.

PR #191, opened 2021-10-09
setup coverage

jyooru:
This PR configures [codecov](https://codecov.io) to track coverage for this project.

It has been configured so that:
- coverage uploads are tagged with matrix variables, `os` and `python-version`
- if coverage goes below 50%, codecov will add a fail status to a PR (this was what was previously set in noxfile.py)
- the following is ignored:
  - noxfile.py
  - code in a `if __name__ == "__main__"` block
  - lines/blocks commented with `# pragma: no cover` - [examples](https://coverage.readthedocs.io/en/coverage-4.0.3/excluding.html#excluding-code-from-coverage-py)

Feel free to change anything, and let me know if you want any configuration changes.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,4 +2,4 @@ coverage:
   status:
     project:
       default:
-        target: 50%
+        target: 90%


bee-san:
This is very cool, thanks! :)

PR #202, opened 2021-10-10
refactor tests/test_regex_identifier.py

jyooru:
**⚠ Pull Requests not made with this template will be automatically closed 🔥**

# Prerequisites
- [x] Have you read the documentation on contributing? https://github.com/bee-san/pyWhat/wiki/Adding-your-own-Regex

# Why do we need this pull request?

Currently, the tests files are quite long. I've attempted moving the test data into `regex.json` and using `pytest.mark.parmaterize` to "generate" tests using the test data now in `regex.json`. This makes it much easier to maintain `test_regex_identifier.py` as to change all `_assert_match_first_item` tests, you only have to change one test:

```py
@pytest.mark.parametrize(
    "name,match",
    [
        (regex["Name"], match)
        for regex in database
        for match in regex.get("Matches", [])
    ],
)
def test_regex_match(name: str, match: str):
    res = r.check([match])
    _assert_match_first_item(name, res)
```

# What [GitHub issues](https://github.com/bee-san/pyWhat/issues) does this fix?

Related to #199
Related to #195  

Contributions to this PR are welcome!

PR #216, opened 2021-10-17
Added loads of test candidates and updated discord webhook regex

IlluminatiFish:
# Prerequisites

Better webhook url regex, to allow other subdomains such as ptb & canary which are actively used as well as allowing for the detection of webhooks using the old domain discordapp.com and allowing for the matching of webhook urls that include the api version after the api path and more distinct expressions for webhook tokens and webhook ids, also allowing for webhook urls that have the scheme of http

Expands upon changes made in #173 and continuation of #208 due to merge conflicts with new regex test system introduced via #202

# Copy / paste of output
Couldn't get the script to work so I could not test the changes to the regex i've made with the regex however I have tested it on its own in python 3.10 as well as using https://pythex.org

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+ignore = E501,W503
+max-line-length = 88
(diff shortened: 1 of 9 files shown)


bee-san:
your tests fail :(

PR #220, opened 2021-10-20
Use the Flake8 Pytest plugin

SkeletalDemise:

# Prerequisites

Adds the Flake8 Pytest plugin and removes the CI Pytest CLI version. This makes it like the other linters and it's easier for contributors to run.

N/A

# Copy / paste of output
N/A

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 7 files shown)


SkeletalDemise:
>>it's easier for contributors to run.

>Why?

Because now contributors can just run `pytest` to run all the tests including Flake8 instead of having to install Flake8 and run it separately with the CLI.

PR #228, opened 2021-11-04
feat: UPS tracking numbers

P403n1x87:

# Prerequisites

* This PR adds the regex for UPS tracking numbers

N. A.

# Copy / paste of output

```console
❯ pywhat 1Z123CAB9912345678    
Matched on: 1Z123CAB9912345678
Name: UPS Tracking Number
Link:  https://www.ups.com/track?tracknum=1Z123CAB9912345678
```


tests/test_regex_identifier.py
```diff
@@ -91,3 +91,10 @@ def test_match_description(match: str, description: str):
 )
 def test_match_exploit(match: str, exploit: str):
     assert exploit in r.check([match])[0]["Regex Pattern"]["Exploit"]
+
+
+def test_ups_tracking():
+    tracking = "1Z123ABC0012345678"
+    (match,) = r.check([tracking])
+    assert match["Matched"] == tracking
+    assert match["Regex Pattern"]["Name"] == "UPS Tracking Number"
```
piatrashkakanstantinass, line 100: Remove this.

pywhat/Data/regex.json
```diff
@@ -1309,6 +1309,23 @@
          ]
       }
    },
+   {
+      "Name": "UPS Tracking Number",
+      "Regex": "^(1Z[0-9A-Z]{6}[0-9]{2}[0-9]{8})$",
+      "plural_name": false,
+      "Description": null,
+      "Rarity": 1,
```
piatrashkakanstantinass, line 1317: I would say that rarity should be lowered.

pywhat/Data/regex.json
```diff
@@ -1309,6 +1309,23 @@
...
+      "URL": "https://www.ups.com/track?tracknum=",
+      "Tags": [
+         "Tracking"
+      ],
```
piatrashkakanstantinass, line 1321: We have some tags, like `Identifiers` and `Bug Bounty` that could fit here. Maybe some others.

piatrashkakanstantinass (changes requested):

pywhat/Data/regex.json
```diff
@@ -1309,6 +1309,23 @@
...
+      "Examples": {
+         "Valid": [
+            "1Z123ABC0012345678"
```
bee-san, line 1324: More examples taken from https://www.quora.com/What-is-an-example-of-a-UPS-tracking-number

```suggestion
            "1Z123ABC0012345678",
            "1Z9999999999999999",
            "1Z999AA10123456784"
```

pywhat/Data/regex.json
piatrashkakanstantinass: Something around 0.3

bee-san: Why 0.3? I'd say higher like 0.5 or 0.6 because:
1. The string has to start with `1Z`
2. It needs 7 chars `0-9A-Z`
3. It has exactly 2 numbers
4. It has 8 numbers

Also, can we make it:

```diff
- ^(1Z[0-9A-Z]{6}[0-9]{2}[0-9]{8})$
- + ^(1Z[0-9A-Z]{6}[0-9]{10})$
```

?

piatrashkakanstantinass: I think 0.4 or 0.5. And yes, regex should be changed.

P403n1x87: The idea of the 2+8 split is because the first 2 digits in this group represent a service indicator code and perhaps it could be captured and handled in the future.

P403n1x87: Aside: I wonder if the "rarity" could be estimated more reliably through some entropy-based measure 🤔

bee-san: > service indicator code

We have precedence for this called sub-categories. See the Mastercard / Phone Numbers regex. I am not sure it'll work on data in the middle of the regex, we may need to change the code for that :)  

> Aside: I wonder if the "rarity" could be estimated more reliably through some entropy-based measure 🤔

Probably! Currently I am estimating it based on what I see when people post this:
Screenshot 2021-11-04 at 16 09 24

And also whether we have any false positives.

pywhat/Data/regex.json
```diff
@@ -1309,6 +1309,23 @@
          ]
       }
    },
+   {
+      "Name": "UPS Tracking Number",
```
bee-san, line 1313: May I suggest United States Postal Service (UPS) Tracking Number? Our target audience may not be aware that UPS is a thing that exists 😄

pywhat/Data/regex.json
piatrashkakanstantinass: @P403n1x87 You can use subcategories with regex method for that.

pywhat/Data/regex.json
```diff
@@ -1753,6 +1753,28 @@
          "Invalid": []
       }
    },
+   {
+      "Name": "United States Postal Service (UPS) Tracking Number",
+      "Regex": "^(1Z[0-9A-Z]{6}[0-9]{2}[0-9]{8})$",
```
piatrashkakanstantinass, line 1758:
```suggestion
      "Regex": "^(1Z[0-9A-Z]{6}[0-9]{10})$",
```

piatrashkakanstantinass (changes requested):

piatrashkakanstantinass (approved):

PR #230, opened 2021-11-06
fix: improved URL regex

amadejpapez:

## Prerequisites

## Why do we need this pull request?
This should fix a few issues we were seeing with URLs. I have went through the regex and modified some parts. There may still be some cases but with this changes I saw a lot better results.

Also added more `Examples` and `https://www.google.com` now matches fully.

I have written an explanation for regex from start of the URL till the end to make it easier and quicker to review. Also give feedback, so it can get even better. :)

```(?i)(?:(?:https?|ftp):\/\/)?(?:\S+:\S+@)?(?:[a-z0-9-_~]+\.)*[a-z0-9-]{1,62}\.(?:COM|IO|BLOG|ORG|TECH)(?::\d{2,5})?(?:\/[a-z0-9-_~.]+)*(?:[?#]\S*)*\/?```

- ```http```/```https```/```ftp``` is still optional at the beggining.
- Subdomains can contain ```[a-z0-9-_~]``` with ```.``` in-between. Previously this part was matched as whole which caused something like ```wwww.....google.com``` to be valid.
- Domain name can contain ```[a-z0-9-]``` and is 1-62 characters long.
- Valid TLD from our list.
- There can be a port number specified.
- Path can contain ```[a-z0-9-_~.]``` with ```/``` in-between.
- If there is ```?``` or ```#``` characters it basically matches to everything after it until there is a space or line break. I do not think this characters can get any more limited.

## What [GitHub issues](https://github.com/bee-san/pyWhat/issues) does this fix?
* Related to #117

## Copy / paste of output

Please copy and paste the output of PyWhat with your new addition using an example that tests this addition below:

```console

```


bee-san (approved):

bee-san (approved):

piatrashkakanstantinass (changes requested):
Please change url generation script

amadejpapez:
> Please change url generation script

What change is needed? Regex is no longer hard-coded in there.

piatrashkakanstantinass:
> > Please change url generation script
> 
> What change is needed? Regex is no longer hard-coded in there.

Oh, that is great!

PR #254, opened 2022-10-25
Fix example invocation with tags in help string

### LINKED ISSUE #253: Passed tags are not valid for example invocation from --help string
**Describe the bug**
Help docstring should be showing working example of how to use pyWhat - in this particular case use valid tags.

**To Reproduce**
```sh
$ what --rarity 0: --include "credentials, username, password" --exclude "aws, credentials" 'James:SecretPassword'
Passed tags are not valid.
You can check available tags by using: 'pywhat --tags'
```

```sh
what --tags | grep -i credentials
Credentials
```

```sh
$ what --tags | grep -i username
$
```

```sh
$ what --tags | grep -i password
$ 
```

**Expected behavior**
Example should return some result, i.e. found something.

Docstring should be showing working example of how to use pyWhat.

**Additional context**
* pyWhat version: 5.1.0+ (main branch)
* python version: 3.9.15
* OS: Linux

mmtj:
This PR updates the help string for example with tags - use valid and currently available tags.

I am not that familiar with all tags and regexes, so it is quite possible that provided example does not necessary shows the best capabilities of pyWhat. I was aiming to alter the example to provide at least some match, instead of warning about incorrect invocation.

So feel free to suggest more meaningful example :-)

## Prerequisites

## Why do we need this pull request?
* Example invocation should lead to at least some result, so user might get familiar with intended usage. IMO it should not show warning about incorrect invocation.

## What [GitHub issues](https://github.com/bee-san/pyWhat/issues) does this fix?
* Fixes #253 

## Copy / paste of output

Please copy and paste the output of PyWhat with your new addition using an example that tests this addition below:

```console
$ what --rarity 0: --include "credentials" --exclude "aws" 'James:SecretPassword'
Matched on: James:SecretPassword
Name: Key:Value Pair
```

### DIFF
--- a/pywhat/what.py
+++ b/pywhat/what.py
@@ -234,7 +234,7 @@ def main(**kwargs):
 
         * what --rarity 0.6: 'myEmail@host.org'
 
-        * what --rarity 0: --include "credentials, username, password" --exclude "aws, credentials" 'James:SecretPassword'
+        * what --rarity 0: --include "credentials" --exclude "aws" 'James:SecretPassword'
 
         * what -br 0.6: -be URL '123myEmail@host.org456'
 


bee-san (approved):

bee-san:
Thanks so much!!

bee-san:
Thanks so much!!

mmtj:
You are welcome, I'm glad I could help :-)

